PR #13, opened 2021-01-28
Implement support for setting specific array indices

### LINKED ISSUE #4: Implement support for setting specific array indices
Add support for setting a specific element of an array

irwineffect:
FIxes #4

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,20 +59,14 @@ macro_rules! impl_array {
             ) -> Result<(), Error> {
                 if let Some(next) = topic_parts.next() {
                     // Parse what should be the index value
-                    let i: usize = serde_json_core::from_str(next)?.0;
+                    let i: usize = serde_json_core::from_str(next).map_err(|_| Error::BadIndex)?.0;
 
-                    // There should not be any more topic parts after the index
-                    if topic_parts.peek().is_some() {
-                        return Err(Error::NameTooLong);
+                    if i >= self.len() {
+                        return Err(Error::BadIndex)
                     }
 
-                    if i < self.len() {
-                        self[i] = serde_json_core::from_slice(value)?.0;
-                        Ok(())
-                    }
-                    else {
-                        Err(Error::BadIndex)
-                    }
+                    self[i].string_set(topic_parts, value)?;
+                    Ok(())
                 }
                 else {
                     let data: [T; $N] = serde_json_core::from_slice(value)?.0;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,6 +10,7 @@ pub enum Error {
     NameTooLong,
     NameTooShort,
     Deserialization(serde_json_core::de::Error),
+    BadIndex,
 }
 
 impl From<serde_json_core::de::Error> for Error {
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -56,12 +57,28 @@ macro_rules! impl_array {
                 mut topic_parts: core::iter::Peekable<core::str::Split<char>>,
                 value: &[u8],
             ) -> Result<(), Error> {
-                if topic_parts.peek().is_some() {
-                    return Err(Error::NameTooLong);
+                if let Some(next) = topic_parts.next() {
+                    // Parse what should be the index value
+                    let i: usize = serde_json_core::from_str(next)?.0;
```
ryan-summers, line 62: Instead of ? on this error, can we catch it and return a `miniconf::Error::InvalidIndex` or something similar? I think a deserialization error would be less intuitive if someone forgot the index.

src/lib.rs
```diff
@@ -56,12 +57,28 @@ macro_rules! impl_array {
...
+
+                    // There should not be any more topic parts after the index
+                    if topic_parts.peek().is_some() {
```
ryan-summers, line 65: This isn't necessarily true. What if it's an array of structures, each of which implement `StringSet`?

src/lib.rs
```diff
@@ -56,12 +57,28 @@ macro_rules! impl_array {
...
+                        return Err(Error::NameTooLong);
+                    }
+
+                    if i < self.len() {
```
ryan-summers, line 69: Would prefer 
```
if i >= self.len() {
    return Err(Error::BadIndex);
}

self[i] = serde_json_core::from_slice(value)?.0;
Ok(())
```

to make the correct path clear

src/lib.rs
```diff
@@ -56,12 +57,28 @@ macro_rules! impl_array {
...
+                        self[i] = serde_json_core::from_slice(value)?.0;
```
ryan-summers, line 70: In lieu of comments above, I think this needs to call `self[i].set_string(topic_parts, value)`

ryan-summers (changes requested):

ryan-summers:
Also, can you add a test case for this?

src/lib.rs
```diff
@@ -59,20 +59,14 @@ macro_rules! impl_array {
             ) -> Result<(), Error> {
                 if let Some(next) = topic_parts.next() {
                     // Parse what should be the index value
-                    let i: usize = serde_json_core::from_str(next)?.0;
+                    let i: usize = serde_json_core::from_str(next).map_err(|_| Error::BadIndex)?.0;
```
ryan-summers, line 62: I have found it's simpler to use `.or(Err(Error::BadIndex))` instead of `map_err()` personally.

tests/arrays.rs
```diff
@@ -0,0 +1,72 @@
...
+    struct S {
+        a: [Inner; 3],
+    }
+
+    let mut s = S::default();
+
+    let field = "a/1/b".split('/').peekable();
+
+    s.string_set(field, "7".as_bytes()).unwrap();
+
```
ryan-summers, line 64: Add a test for setting a field that doesn't exist in the struct as well

tests/arrays.rs
```diff
@@ -0,0 +1,72 @@
...
+        a: [u8; 3],
+    }
+
+    let mut s = S::default();
+
+    let field = "a/1".split('/').peekable();
+
+    s.string_set(field, "7".as_bytes()).unwrap();
+
+    assert_eq!([0, 7, 0], s.a);
```
ryan-summers, line 44: Add a test requesting an out-of-bounds index

ryan-summers (changes requested):

ryan-summers (approved):

PR #14, opened 2021-01-30
Adding MQTT-based interface

### LINKED ISSUE #5: Add an MQTT-based settings interface
Provide a generic implementation of the settings manager on top of the minimq MQTT interface.

**Note**: This involves porting https://github.com/quartiq/stabilizer/pull/242/files#diff-07a695d6c0ee371b1c2dba3a29717f2235b6a8ae61880e086e5f607186b3edd8 to this repository

ryan-summers:
This PR adds an MQTT-based settings interface and closes #5.

There is also an integration test added that can run if a broker (e.g. mosquitto) is running on the local machine.

### DIFF
--- a/src/mqtt_interface.rs
+++ b/src/mqtt_interface.rs
@@ -1,25 +1,25 @@
-use minimq::{
-    embedded_nal::{TcpStack, IpAddr, Ipv4Addr},
-    QoS, Property, MqttClient
-};
-use heapless::{String, consts};
-use core::fmt::Write;
 use super::StringSet;
+use core::fmt::Write;
+use heapless::{consts, String};
+use minimq::{
+    embedded_nal::{IpAddr, TcpStack},
+    MqttClient, Property, QoS,
+};
 
 #[derive(Debug)]
-pub enum Error<S: TcpStack> where S::Error: core::fmt::Debug {
+pub enum Error<E: core::fmt::Debug> {
     /// The provided device ID is too long.
     IdTooLong,
 
     /// MQTT encountered an internal error.
-    Mqtt(minimq::Error<S::Error>),
+    Mqtt(minimq::Error<E>),
 
     /// The network stack encountered an error.
-    Network(S::Error)
+    Network(E),
 }
 
-impl<S: TcpStack> From<minimq::Error<S::Error>> for Error<S> {
-    fn from(err: minimq::Error<S::Error>) -> Self {
+impl<E: core::fmt::Debug> From<minimq::Error<E>> for Error<E> {
+    fn from(err: minimq::Error<E>) -> Self {
         match err {
             minimq::Error::Network(err) => Error::Network(err),
             other => Error::Mqtt(other),
@@ -42,7 +42,10 @@ pub enum Action {
 //
 // # Returns
 // The string - otherwise, an error indicating the generated string was too long.
-fn generate_topic<'a, 'b>(device_id: &'a str, topic: &'b str) -> Result<String<minimq::consts::U128>, ()> {
+fn generate_topic<'a, 'b>(
+    device_id: &'a str,
+    topic: &'b str,
+) -> Result<String<minimq::consts::U128>, ()> {
     let mut string: String<consts::U128> = String::new();
     write!(&mut string, "{}/{}", device_id, topic).or(Err(()))?;
     Ok(string)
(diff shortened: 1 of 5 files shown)


src/mqtt_interface.rs
```diff
@@ -0,0 +1,258 @@
...
+                    .ok();
+                return;
+            }
+
+            // Process the command
+            let response = match split.next() {
+                Some("settings") => {
+                    // Handle settings failures
+                    let mut response: String<consts::U512> = String::new();
+                    match self.settings.string_set(split.peekable(), message) {
```
irwineffect, line 186: This library is doing some pretty large stack allocations between these temporary buffers and recursive calls to `string_set()`, I wonder if we should do some analysis and mention it in the README? 

It's probably fine for RTIC with its "one stack" design, but threaded RTOSs can often have pretty small stacks by default, 1KB or even less.

src/mqtt_interface.rs
```diff
@@ -0,0 +1,258 @@
...
+        // If we are not yet subscribed to the necessary topics, subscribe now.
+        if !self.subscribed && client.is_connected()? {
+            client.subscribe(&self.settings_topic, &[])?;
+            client.subscribe(&self.commit_topic, &[])?;
+            self.subscribed = true;
+        }
+
+        let mut result = Action::Continue;
+
+        let result = match client.poll(|client, topic, message, properties| {
```
irwineffect, line 132: This is kinda confusing have multiple `result` here, especially because the first `result` is used inside the match statement that calculates the second `result`.

I wonder if we could make the logic of this match statement more clear by creating a separate function or something, it's nearly 100 lines.

src/mqtt_interface.rs
```diff
@@ -0,0 +1,258 @@
...
+}
+
+/// An interface for managing MQTT settings.
+pub struct MqttInterface<T, S>
+where
+    T: StringSet,
+    S: TcpStack,
+{
+    // TODO: Allow the user to specify buffer size.
+    client: Option<MqttClient<minimq::consts::U256, S>>,
```
irwineffect, line 58: It seems like `client` is never `None` except within the body of each function, so why wrap in an `Option`? Is it to protect against the possibility of two different functions being called at the same time? Basically panic if the user uses the library wrong?

src/mqtt_interface.rs
jordens: Right. Stack usage here and in minimq in addition to the memory usage in the nal and smoltcp sounds like it might be a bit on the high side.

src/mqtt_interface.rs
```diff
@@ -0,0 +1,259 @@
...
+    /// * `id` - The ID for uniquely identifying the device. This must conform with MQTT client-id
+    ///          rules. Specifically, only alpha-numeric values are allowed.
+    /// * `broker` - The IpAddress of the MQTT broker.
+    /// * `settings` - The initial settings of the interface.
+    ///
+    /// # Returns
+    /// A new `MqttInterface` object that can be used for settings configuration and telemtry.
+    pub fn new(
+        stack: S,
+        id: &str,
```
ryan-summers, line 86: We're treating the client ID and the prefix ID identically here, but there are different rules that MQTT enforces for these values. Perhaps we need to allow a client-id and a prefix.

It may make more sense to allow the user to supply an MqttClient, as this would allow them to specify an arbitrary ID, broker, etc. Then, the prefix of the device could be selectable.

src/mqtt_interface.rs
```diff
@@ -0,0 +1,258 @@
...
+                        response_topic,
+                        "No ID speciifed".as_bytes(),
+                        QoS::AtMostOnce,
+                        &[],
+                    )
+                    .ok();
+                return;
+            }
+
+            if id.unwrap() != self.id {
```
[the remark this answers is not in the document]
ryan-summers: I had originally structured it this way, but you get a massive `if let` cascade that nests pretty heavily, so I went for this approach instead.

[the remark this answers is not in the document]
ryan-summers: Refactored and this should be resolved now.

src/mqtt_interface.rs
ryan-summers: This is actually to get around rust's ownership rules. The issue here is that `minimq::MqttClient` takes a `FnMut` closure to process incoming messages. However, that results in us mutably borrowing the `self.client`. This means that we aren't allowed to then use `self` mutably _within_ that closure. The solution is to store the `client` in an `Option` and move it in and out of `self` as needed - I'm open to other ideas though.

src/mqtt_interface.rs
ryan-summers: I've refactored this to take significantly less stack. Note that the recursive calls shouldn't be adding very much overhead in each call, since they're extremely simple functions. Additionally, there's not much overhead using this as a layer between MQTT and the application, as it's only a single function call of abstraction. The network servicing routines are not called within these scopes, so that's not of concern here.

ryan-summers:
Will need to merge https://github.com/quartiq/minimq/pull/26 before this can land. Currently fighting some GHA issues

src/mqtt_interface.rs
irwineffect: I see....Hmm I was thinking a Cell might help there, but that would look really similar to what you already have.

irwineffect (approved):

PR #16, opened 2021-02-15
Updating to minimq 0.2

ryan-summers:


ryan-summers:
I just released 0.2 of minimq on crates.io, so we can leverage all of the updates directly from crates.io, which allows us to publish this crate when ready.

jordens (approved):

PR #36, opened 2021-03-07
Adapt to minimq with embedded-time integration for keepalive

dnadlinger:
Dependency needs to be updated after the minimq PR has been merged.

Issue #35, opened 2021-03-07
Accept minimq messages rather than owning the client


dnadlinger:
In many `miniconf` applications, it might be desirable to expose an MQTT interface beyond the auto-generated settings interface (e.g. to trigger sampling of some measurements, query fast-changing status information, etc.).

Currently, `MqttInterface` takes over the `minimq::MqttClient`, so this isn't possible.

The easiest option would probably be to have a `try_process` method instead that the client code can call from the MqttClient message callback, so messages outside the miniconf tree can be handled by application-specific code.

dnadlinger:
(This also avoids piping all the network stack/… through MqttInterface, which is a bit clunky.)

dnadlinger:
Okay, actually it doesn't remove the need for all the network stack type parameters, as there is no general (non-generic, or only in the error type) MqttInterface trait.

In that instance, we might as well add an extra callback parameter to `update` to which unknown messages are forwarded.

ryan-summers:
Just to elaborate - there's also a `miniconf::MqttInterface::client()` function which accepts a closure. That closure is provided the MQTT client object, so you can still perform arbitrary actions directly with the underlying client. This API was intended to allow these types of use-cases that you're referring to.

dnadlinger:
But what I need to do is handle messages other than the ones expected by Miniconf. How would that work with the `client()` API?

ryan-summers:
That's a valid point - I'll have to think about the architecture here to see if that would make sense

ryan-summers:
The more I've thought about this, the more I'm coming under the opinion that owning the client is the wrong API. It makes it really difficult to use MQTT for anything besides miniconf.

I'm going to work on a refactor to pull MQTT ownership out of miniconf. Theoretically, miniconf doesn't even depend on MQTT - it just depends on some topic-data pairing that serializes data with JSON.

PR #37, opened 2021-03-07
Add update_or_process() for unexpected message topics

dnadlinger:
Includes #36.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,11 @@ derive_miniconf = { path="derive_miniconf" }
 serde-json-core = "0.2.0"
 serde = { version = "1.0.120", features = ["derive"], default-features = false }
 heapless = ">=0.6.1"
-minimq = "0.2"
+embedded-time = "0.10.1"
+
+[dependencies.minimq]
+git = "https://github.com/dnadlinger/minimq.git"
+branch = "feature/keepalive"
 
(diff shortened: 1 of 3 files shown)


dnadlinger:
This is an alternative to the approach proposed in #35, whereby miniconf still owns the minimq client, but accepts an unexpected topic handler as a closure that can be used to respond to non-settings topics. This is still a bit messy, as the client code also effectively needs to duplicate the (re-)subscription handling code.

I'm not sure what the best design is here, but see https://github.com/dnadlinger/stabilizer/commit/de189c2f5c9469fc4cca3ff5afaa33db1a8ba6db for a quickly hacked-together use case of this (adds a "query" topic which can be used to read back a continuously updating value).

ryan-summers:
Closing. MQTT functionality has been removed from this library - if we want this functionality in Stabilizer, we can add it there directly if desired, although I believe the re-architecture makes this redundant. Let me know if you have any issues!

dnadlinger:
Was about to do that just now, literally was already on the page. :)

Adding a hook to stabilizer's `net` to allow people to adding additional topics in additional binaries without changes to the shared code would probably still be interesting, though.

ryan-summers:
By the way - thanks for the feedback - definitely helped with figuring out the right re-architecture :)

PR #38, opened 2021-03-16
Attach any request correlation data to responses

dnadlinger:
This behaviour is suggested in a (non-normative) comment in the MQTT 5.0
spec, and allows requesters to keep track of multiple in-flight operations.


ryan-summers (approved):
LGTM! Thanks for the PR - I didn't have notifications set up properly for this repository, so sorry for the delay.

PR #39, opened 2021-04-22
Removing minimq MQTT client ownership

### LINKED ISSUE #35: Accept minimq messages rather than owning the client

ryan-summers:
This PR removes MQTT client ownership from `miniconf` in favor of a generic API.

This has the following benefits:
* MQTT can now be used for things other than `miniconf`
* Unsupported messages are no longer silently dropped (fixes #35)
* Any MQTT client (or in fact, any other similar protocol) may be used for miniconf


src/interface.rs
```diff
@@ -0,0 +1,131 @@
+use super::{Error, Miniconf};
+use core::fmt::Write;
+use heapless::{consts, String};
+
+/// Represents an incoming MQTT packet.
+pub struct Message<'a> {
```
ryan-summers, line 6: Let's add first-class support for minimq for the `Message` and `Response` properties - specifically, ```
impl From<(&[u8], &[minimq::Property])> for Message { ... }
impl Response {
    to_minimq(&self) -> (&[u8], &[minimq::Property));
}
```

irwineffect (changes requested):
Could we add some example code to the docs showing how an application now uses the interface? (Assuming using MQTT)

irwineffect (approved):
Looks good!

PR #42, opened 2021-07-14
Adding docs for errors

ryan-summers:
This PR adds documentation for each error code.

This will allow stabilizer docs to defer to cargo-doc output for error codes returned via the miniconf run-time settings interface.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,13 +5,38 @@ pub use serde_json_core;
 
 pub use derive_miniconf::{Miniconf, MiniconfAtomic};
 
+/// Errors that occur during settings configuration
 #[derive(Debug, PartialEq)]
 pub enum Error {
+    /// The provided name wasn't found in the structure.
+    ///
+    /// Double check the provided name to verify that it's valid.
     NameNotFound,
+
+    /// The provided name was valid, but there was trailing data at the end.
+    ///
+    /// Check the end of the name and remove any excess characters.
     NameTooLong,
+
+    /// The provided name was valid, but did not specify a value fully.
+    ///
+    /// Double check the ending and add the remainder of the name.
     NameTooShort,
+
+    /// The name provided is attempting to configure a portion of the structure, but the structure
+    /// must be updated all at once.
+    ///
+    /// Refactor the request to configure the entire structure at once.
     AtomicUpdateRequired,
+
+    /// The value provided for configuration could not be deserialized into the proper type.
+    ///
+    /// Double check the serialized data and correct any errors.
     Deserialization(serde_json_core::de::Error),
+
+    /// When indexing into an array of names, the index provided was out of bounds.
+    ///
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -5,13 +5,38 @@ pub use serde_json_core;
 
 pub use derive_miniconf::{Miniconf, MiniconfAtomic};
 
+/// Errors that occur during settings configuration
 #[derive(Debug, PartialEq)]
 pub enum Error {
+    /// The provided name wasn't found in the structure.
```
jordens, line 11: Shouldn't this be `key` instead of `name` for consistency everywhere?
And this should all be present tense.

src/lib.rs
```diff
@@ -5,13 +5,38 @@ pub use serde_json_core;
...
+
+    /// The name provided is attempting to configure a portion of the structure, but the structure
+    /// must be updated all at once.
+    ///
+    /// Refactor the request to configure the entire structure at once.
     AtomicUpdateRequired,
+
+    /// The value provided for configuration could not be deserialized into the proper type.
+    ///
+    /// Double check the serialized data and correct any errors.
```
jordens, line 34:
```suggestion
    /// Check that the serialized data is valid JSON and of the correct type.
```

src/lib.rs
```diff
@@ -5,13 +5,38 @@ pub use serde_json_core;
...
+
+    /// The provided name was valid, but did not specify a value fully.
+    ///
+    /// Double check the ending and add the remainder of the name.
     NameTooShort,
+
+    /// The name provided is attempting to configure a portion of the structure, but the structure
+    /// must be updated all at once.
+    ///
+    /// Refactor the request to configure the entire structure at once.
```
jordens, line 29:
```suggestion
    /// Refactor the request to configure the surrounding structure at once.
```

jordens (changes requested):

src/lib.rs
ryan-summers: This is using `name` in the docs because the error codes currently use the "name" terminology. I have no preference either way - technically, it refers to a variable name in the struct, but it's also a key to accessing a struct member, so I think both are logically correct.

jordens: Ok. But then shouldn't the stabilizer docs also use "name" to refer to the miniconf keys?

jordens: Or "path"?

jordens (approved):

src/lib.rs
ryan-summers: I'm honestly of the opinion here to rename all the errors to refer to it as a path, as it _is_ a path-like object to get to the member.

jordens: Agreed.
Matches https://github.com/quartiq/stabilizer/blob/master/miniconf.py

PR #44, opened 2021-08-05
Adding python package / utility

ryan-summers:
This PR adds a python package fore interfacing with Miniconf. This is moving the `stabilizer` `miniconf.py` to live here so that it can be managed alongside `miniconf` and  be published to PyPi

TODO:
- [x] Set up Python style checks


miniconf-py/setup.py
```diff
@@ -0,0 +1,21 @@
...
+Author: Vertigo Designs, Ryan Summers
+
+Description: Setup file for Miniconf packaging.
+"""
+from setuptools import setup, find_packages
+
+# Load the version string from the version file.
+execfile('miniconf/version.py')
+
+setup(name='miniconf-rs',
```
jordens, line 12: Maybe
```suggestion
setup(name='miniconf-py',
```
?

miniconf-py/README.md
```diff
@@ -0,0 +1,6 @@
+# Miniconf Python Utility
+
+This directory contains a Python package for interacting with Miniconf utilities.
+
+## Installation
+Run `pip install .` from this directory to install the `miniconf` package.
```
jordens, line 6: Is there a way to make this `pip install git+https://...` compatible? That would make the installation nicer.
We might just want to split the repo.

jordens (changes requested):
A couple suggestions, no strong opinion though.

miniconf-py/setup.py
ryan-summers: This is the name it gets published on `pypi` as - I'd prefer if it was just `miniconf`, but that's taken, so I was opting for `miniconf-rs` so that users would be able to easily correlate it to the rust `miniconf` package. Open to renaming suggestions though. I think `miniconf-py` is a bit difficult to discern if you're just looking at pypi.

miniconf-py/README.md
ryan-summers: I would assume it already is? You should be able to do `pip install git+https://github.com/quartiq/miniconf@feature/python-utility#subdirectory=miniconf-py` (with the absence of branch once merged) as well (once I fix the build)

ryan-summers: Added a note to the README on how to do this.

miniconf-py/setup.py
jordens: Then miniconf-mqtt as this is an impl on one specific transport.

jordens (approved):
miniconf-mqtt is also a good option imo.
Either is fine.

PR #46, opened 2021-08-06
Fixing main utility

ryan-summers:
This PR fixes the `__main__` utility, which was inadvertently broken in the move-over as we migrated to using exceptions on error.


miniconf-py/miniconf/__main__.py
```diff
@@ -47,11 +47,9 @@ async def configure_settings():
...
-                                               not args.no_retain)
-            print(f'{path}: {response}')
-            if response['code'] != 0:
-                return response['code']
+            await interface.command(path, json.loads(value),
+                                           not args.no_retain)
+            print(f'{path}: OK')
         return 0
 
     sys.exit(loop.run_until_complete(configure_settings()))
```
jordens, line 55: Since the exit code is now set elsewhere on error, then we should also:

```suggestion
    loop.run_until_complete(configure_settings())
```

And remove the sys import.

jordens (changes requested):

jordens:
And while we are at it, if the python module is `miniconf-mqtt`, then the directory should preferrably be the module name (or `py/miniconf-mqtt` if you prefer to disambiguate the language namespaces).

jordens (approved):

PR #49, opened 2021-08-11
Preparing library for release

ryan-summers:
This PR cleans up Miniconf for initial release. It also does the following:
* Add support for using const-generics to implement `Miniconf` for array types
* Added an async example for running Miniconf on-host
* Revamps docs to include more helpful information


jordens (approved):
Nice

PR #50, opened 2021-08-20
Feature/minimq update

ryan-summers:
This PR updates `miniconf` to utilize an updated version of minimq that uses ping timeouts.

TODO:
- [x] Release patch to minimq with bumped version

### DIFF
--- a/examples/mqtt.rs
+++ b/examples/mqtt.rs
@@ -18,7 +18,7 @@ struct Settings {
 
 async fn mqtt_client() {
     // Construct a Minimq client to the broker for publishing requests.
-    let mut mqtt: Minimq<_, _, 256, 3> = Minimq::new(
+    let mut mqtt: Minimq<_, _, 256, 1> = Minimq::new(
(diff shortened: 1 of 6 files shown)


examples/mqtt.rs
```diff
@@ -17,11 +18,16 @@ struct Settings {
 
 async fn mqtt_client() {
     // Construct a Minimq client to the broker for publishing requests.
-    let mut mqtt: Minimq<_, 256> =
-        Minimq::new("127.0.0.1".parse().unwrap(), "tester", Stack::default()).unwrap();
+    let mut mqtt: Minimq<_, _, 256, 3> = Minimq::new(
```
jordens, line 21: Do we need more than 1 (or even 0?) in flight QoS messages for miniconf?

```suggestion
    let mut mqtt: Minimq<_, _, 256, 0> = Minimq::new(
```

ryan-summers: We cannot use a size of zero because of the underlying storage contains in minimq, but we can use zero.

Issue #10, opened 2021-01-28
Add setting re-publication (readback) support


ryan-summers:
~Add support for reading back a staged settings value~

Add support for publishing active settings. Those settings may differ from the MQTT view for several reasons:

* Rounding due to value conversion (e.g. DDS FTW words)
* Custom validation hooks, value boundaries
* Application changes
* No data in MQTT available yet (because it was never set)

jordens:
I'm not sure pull readback would be a prime usecase. It would be better to generally ensure that the MQTT view of the active settings is consistent with the device reality. In that sense the solution would be to publish the settings when they are activated. That also addresses:

* Making the active settings available in MQTT when the device boots and there are no retained settings
* Publishing settings changes due to firmware changes
* Validation/implicit rejection of the settings from the staged ones on commit
* Possible interdependencies of settings (the range of one value depending on another).
* Pivoting to the configured settings tree (#25)

ryan-summers:
To update, the primary use case here is to publish the default, power-on active settings if no data is received over the MQTT interface so that the active configuration is available via the broker. Settings should be published back to the topic that is used to configure them in the first place.

PR #62, opened 2021-11-11
Feature/readback design

irwineffect:
This draft PR is to review the prototype design for implementing #10.

Readback support for a struct is implemented as recursive iterators. Prototype is only partially implemented via proc macros, see examples/readback.rs for remainder of the prototype. 

For the top-level struct, I'm implementing a classic `Iterator` wrapper around the `recursive_iterator()` method. This is convenient for being able to go over the struct with for-loop syntax:
```Rust
for (topic, value) in s.miniconf_iter(&mut iterator_state) {
...
}
```
However, this means that the iterator owns the Settings for as long as the iterator exists.

I also implemented alternative method for iteration that presents a similar interface, but is different from the `Iterator` trait. In the alternative implementation, ownership of the settings is only for the time of the `next()` call, no need to destroy the iterator in order to access the settings.


src/lib.rs
```diff
@@ -169,6 +178,36 @@ macro_rules! impl_single {
...
             }
+
+            // This implementation is the base case for primitives where it will
+            // yield once for self, then return None on subsequent calls.
+            fn recursive_iter<const TS: usize, const VS: usize>(
+                &self,
+                index: &mut [usize],
+                _topic: &mut heapless::String<TS>,
+                value: &mut heapless::String<VS>,
+            ) -> Option<()>
```
ryan-summers, line 189: Instead of pass-by-reference, just have the prototype for `recursive_iter()` construct the `topic` and `value` internally. No reason to pass it as input from what I can see.

Also, `value` should be a `Vec<u8>`. `String` is a UTF-8 encodeded string, and we make no guarantees about the serialization format of the underlying data right now.

ryan-summers: Ah wait, I now see why they're passed as input arguments. That's because we're constantly appending to `topic`. That being said, `value` could be returned as the result instead - I don't see why that needs to be an input argument

ryan-summers: Or alternatively take `value` as a mutable slice reference so users can manage memory for it and we lose a template argument

src/lib.rs
```diff
@@ -76,6 +76,8 @@ pub use serde_json_core;
 
 pub use derive_miniconf::{Miniconf, MiniconfAtomic};
 
+pub use heapless;
```
ryan-summers, line 79: If we could keep `heapless` out of the public API, I think that may be beneficial. That helps ease dependency issues later

examples/readback.rs
```diff
@@ -0,0 +1,145 @@
...
+    let settings_iter = s.miniconf_iter::<128, 10>(&mut iterator_state);
+    for (topic, value) in settings_iter {
+        println!("{} {}", topic, value);
+    }
+
+    println!("\nAlternative Implementation:");
+    let mut value_buffer: String<128> = String::new();
+    let mut topic_buffer: String<128> = String::new();
+
+    let mut iter_state: IterState<Settings, 5> = IterState::new();
```
ryan-summers, line 132: Let's move `value_buffer` and `topic_buffer` into `IterState` so the user doesn't have to explicitly pass them into `next()`.

Ultimately, we want to iterate over them like:
```rust

for (topic, value) in settings_iterator {
    // ... Do stuff, break out when can't do more stuff
}
```

ryan-summers (approved):

PR #63, opened 2021-12-07
Feature/settings validation

### LINKED ISSUE #51: Constraint validation
There should be a way to reject semantically invalid values (instead of just syntactically invalid as now).
This is setting-dependent.

* `f32` bounds (max, min)
* Rounding
* Complicated multi-settings constraints (iir filter coefficients leading to undesired zero/pole positions due to rounding)

ryan-summers:
This PR fixes #51 by adding support for a custom handler to be called whenever settings are updated.

Note: Currently, if a new setting is rejected, it still persists in the cached settings.

### DIFF
--- a/src/mqtt_client/mod.rs
+++ b/src/mqtt_client/mod.rs
@@ -1,3 +1,19 @@
 mod messages;
 mod mqtt_client;
 pub use mqtt_client::MqttClient;
+
+#[non_exhaustive]
+pub enum HandlerResult {
+    /// A settings updated was accepted.
+    UpdateAccepted,
+
+    /// The application of a setting had ramifications on other settings values. In this case, all
+    /// settings will be republished.
+    UpdateSideEffects,
+}
+
+impl From<&()> for HandlerResult {
+    fn from(_: &()) -> Self {
+        HandlerResult::UpdateAccepted
+    }
+}
--- a/src/mqtt_client/mqtt_client.rs
+++ b/src/mqtt_client/mqtt_client.rs
@@ -291,8 +291,8 @@ where
     ///
     /// // let mut client = miniconf::MqttClient::new(...);
     /// client.handled_update(|path, old_settings, new_settings| {
-    ///     if old_settings.threshold > 5 {
-    ///         return Err(());
+    ///     if new_settings.threshold > 5 {
+    ///         return Err("Requested threshold too high");
     ///     }
     ///
     ///     *old_settings = new_settings.clone();
@@ -306,7 +306,7 @@ where
     pub fn handled_update<F, E>(&mut self, handler: F) -> Result<bool, minimq::Error<Stack::Error>>
     where
         F: FnMut(&str, &mut Settings, &Settings) -> Result<(), E>,
-        E: core::fmt::Debug,
+        E: AsRef<str>,
     {
         if !self.mqtt.client.is_connected() {
             // Note(unwrap): It's always safe to reset.
@@ -344,7 +344,7 @@ where
     ) -> Result<bool, minimq::Error<Stack::Error>>
     where
         F: FnMut(&str, &mut Settings, &Settings) -> Result<(), E>,
-        E: core::fmt::Debug,
+        E: AsRef<str>,
     {
         let mut settings = &mut self.settings;
         let mqtt = &mut self.mqtt;
@@ -376,15 +376,17 @@ where
             let message: SettingsResponse =
                 match new_settings.string_set(path.split('/').peekable(), message) {
                     Ok(_) => {
-                        if handler(&path, &mut settings, &new_settings).is_ok() {
-                            // If the update was accepted, store the updated settings into the
-                            // settings structure.
-                            updated = true;
+                        updated = true;
+                        handler(&path, &mut settings, &new_settings).into()
+                    }
+                    err => {
+                        let mut msg = String::new();
+                        if write!(&mut msg, "{:?}", err).is_err() {
+                            msg = String::from("Configuration Error");
                         }
 
-                        SettingsResponse::ok()
+                        SettingsResponse::error(msg)
                     }
-                    err => SettingsResponse::from(err),
                 };
 
             let response = MqttMessage::new(properties, default_response_topic, &message);
@@ -422,7 +424,7 @@ where
     pub fn update(&mut self) -> Result<bool, minimq::Error<Stack::Error>> {
         self.handled_update(|_, old, new| {
             *old = new.clone();
-            Result::<(), ()>::Ok(())
+            Result::<(), &'static str>::Ok(())
         })
     }
 
--- a/src/mqtt_client/mqtt_client.rs
+++ b/src/mqtt_client/mqtt_client.rs
@@ -79,11 +79,19 @@ where
         })
     }
 
-    /// Update the MQTT interface and service the network
+    /// Update the MQTT interface and service the network. Validate any settings changes.
+    ///
+    /// # Args
+    /// * `apply` - A closure called with any updated settings that is used to apply current
+    ///   settings and validate the configuration.
     ///
     /// # Returns
     /// True if the settings changed. False otherwise.
-    pub fn update(&mut self) -> Result<bool, minimq::Error<Stack::Error>> {
+    pub fn update<F, E>(&mut self, mut apply: F) -> Result<(), minimq::Error<Stack::Error>>
+    where
+        F: FnMut(&Settings) -> Result<(), E>,
+        E: core::fmt::Debug,
+    {
         // If we're no longer subscribed to the settings topic, but we are connected to the broker,
         // resubscribe.
         if !self.subscribed && self.mqtt.client.is_connected() {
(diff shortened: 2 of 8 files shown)


src/mqtt_client/mqtt_client.rs
```diff
@@ -321,14 +336,28 @@ where
...
-
-            let message: SettingsResponse = settings
-                .string_set(path.split('/').peekable(), message)
-                .map(|_| {
-                    update = true;
-                })
-                .into();
+            let mut buffer = [0; MESSAGE_SIZE];
+            let message: SettingsResponse =
+                match settings.string_get(path.split('/').peekable(), &mut buffer) {
```
ryan-summers, line 341: I'm not super happy with this implementation. It makes us serialize the setting that is being updated on every change. We could alternatively just require that `Settings: Copy` and make a deep copy of the settings structure. It's unknown which one would be the least overhead.

Additionally, the code flow here is pretty convoluted and could likely be broken out into a separate function or refactored.

src/mqtt_client/mqtt_client.rs
```diff
@@ -321,14 +336,28 @@ where
...
+                        match settings.string_set(path.split('/').peekable(), message) {
+                            Ok(_) => {
+                                updated = true;
+                                let result = handler(settings);
+                                if result.is_err() {
+                                    // Note(unwrap): We just serialized this value, so it should
+                                    // always be valid.
+                                    settings
+                                        .string_set(path.split('/').peekable(), &buffer[..len])
+                                        .unwrap();
```
ryan-summers, line 352: Need to set updated to false if we rejected the settings.

I think we also need to republish the correct value.

tests/validation_failure.rs
```diff
@@ -0,0 +1,107 @@
...
+                }
+
+                return Ok(());
+            })
+            .unwrap()
+        {
+            break;
+        }
+
+        tokio::time::sleep(tokio::time::Duration::from_millis(10)).await;
```
ryan-summers, line 105: Check that `settings.error` has properly been de-asserted.

ryan-summers:
@jordens Would you mind taking a look at the implementation here for some thoughts? I'm trying to answer the following questions:
1. Should settings requests be rejectable? 
2. If so, what should miniconf do on rejection? (e.g. restore settings, republish previous values, more?)

src/mqtt_client/mqtt_client.rs
jordens: Hmm. Yes that doesn't look good.
1. To separate requested settings from actual active setting on mqtt we may want to separate those trees on the topic level.
2. I would usually expect the sequence to be (1) deser of the payload, (2) validation (3) application/update. Maybe 2 and 3 can both be done by the handler.

src/mqtt_client/mqtt_client.rs
```diff
@@ -253,17 +265,27 @@ where
...
+    /// Update the MQTT interface and service the network. Handle any settings changes.
+    ///
+    /// # Args
+    /// * `handler` - A closure called with updated settings that can be used to apply current
+    ///   settings or validate the configuration.
     ///
     /// # Returns
     /// True if the settings changed. False otherwise.
-    pub fn update(&mut self) -> Result<bool, minimq::Error<Stack::Error>> {
+    pub fn handled_update<F, E, T>(
```
jordens, line 280: Ah. It took me five attempts to understand what "handled_update()" means.
Maybe just unify and have `update<F, E, T>(&mut self, handler: Option<F>)`?

src/mqtt_client/mod.rs
```diff
@@ -1,3 +1,19 @@
...
+    /// The application of a setting had ramifications on other settings values. In this case, all
+    /// settings will be republished.
+    UpdateSideEffects,
+}
+
+impl From<&()> for HandlerResult {
+    fn from(_: &()) -> Self {
+        HandlerResult::UpdateAccepted
+    }
+}
```
jordens, line 19: I thought about it a bit more. I think we can skip this. Instead have the `Result::Ok` from the handler be returned from `update() -> Result` to the application. Then the application can decide based on that result if and what to do. We just need to make sure we expose the republication API to the application.
What do you think?

jordens: And we should also give the handler two more bits of information if possible: a ref to the old settings and the path of the update.

src/mqtt_client/mqtt_client.rs
ryan-summers: My onle goal with adding `handled_update()` was that it was a way to maintain backwards compatibility with the previous API.

The proposed API is really unergonomic if you don't have a handler, since Rust will complain that the type `F` cannot be automatically determined.

src/mqtt_client/mod.rs
ryan-summers: I've added additional info to the handler.

However, we can't allow the `update()` to return the result of the handler, since the handler may be called N times due to the MQTT poll loop. We can refactor that in the future, but for now, I've updated it so that the handler has to return Result<(), E>`

src/mqtt_client/mqtt_client.rs
```diff
@@ -48,16 +48,25 @@ mod sm {
     statemachine! {
         transitions: {
             *Initial + Connected = ConnectedToBroker,
-            ConnectedToBroker + Subscribed = IndicatingLife,
-            IndicatingLife + IndicatedAlive / start_republish_timeout = PendingRepublish,
+            ConnectedToBroker + IndicatedLife = PendingSubscribe,
+
+            // After initial subscriptions, we start a timeout to republish all settings that is
+            // reset by any incoming updates.
```
jordens, line 54: I would not reset the timeout on incoming updates. That creates unnecessary dependencies.

src/mqtt_client/mqtt_client.rs
```diff
@@ -252,17 +261,23 @@ where
...
         {
-            self.state
-                .process_event(sm::Events::IndicatedAlive)
-                .unwrap();
+            self.state.process_event(sm::Events::IndicatedLife).unwrap();
         }
     }
 
-    /// Update the MQTT interface and service the network
+    /// Update the MQTT interface and service the network. Handle any settings changes.
```
jordens, line 268:
```suggestion
    /// Update the MQTT interface and service the network. Pass any settings changes to the handler supplied.
```
```suggestion
    /// Update the MQTT interface and service the network. Handle any settings changes.
```

src/mqtt_client/mqtt_client.rs
```diff
@@ -358,8 +386,25 @@ where
         }
     }
 
+    /// Update the settings from the network stack without any specific handling.
+    ///
+    /// # Returns
+    /// True if the settings changed. False otherwise
+    pub fn update(&mut self) -> Result<bool, minimq::Error<Stack::Error>> {
+        self.handled_update(|_, _, _| Result::<(), ()>::Ok(()))
```
jordens, line 394: I would give the handler the task to do the settings copy. That would support more complicated use cases where the desired new settings are more easily computed from the old and not the new and save one of the two settings copying.

```suggestion
        self.handled_update(|_path, old, new| { *old = new; Result::<(), ()>::Ok(()) })
```

src/mqtt_client/mod.rs
jordens: Can't we just exit the loop on error? The unhandled incoming data will still be there (and presumably be handled next time) and the TX buffer is guaranteed to drain (so that there will be no deadlock).

src/mqtt_client/mqtt_client.rs
jordens: Ack. Then maybe worth thinking a bit more about the docstring to make it clear what the function is and does. I think that would help me already.

src/mqtt_client/mod.rs
ryan-summers: As of right now, minimq doesn't provide a mechanism to return an `Error` from inside the closure - the result of processing is unused when considering continuing iteration:
https://github.com/quartiq/minimq/blob/e8a072726ca94efbd2048917218cb2ba24b5c599/src/mqtt_client.rs#L413

I'm not opposed to refactoring minimq to support this behavior - I'll open up an issue.

jordens: Ack. I'll leave it to you whether you want to wait for that in `minimq` and then follow up here or land this here first and then refactor later. But it'll ripple through the stack.

ryan-summers: Actually, after testing it out more, the application can just pass information back via accessing external state in the closure, so it can track any results that way if need be. I actually think the `Result` of this closure should be used for transferring a response back to the original requestee.

For example:
```rust
let mut republish_required = false;
client.handled_update(|path, old, new| {
    if new.threshold > MAX_THRESHOLD {
        return Err("Requested threshold above maximum";
    }

    *old = new.clone();
    Ok(())
}).unwrap()
```

src/mqtt_client/mqtt_client.rs
ryan-summers: I added a complete code example to show the user.

jordens (approved):

PR #65, opened 2021-12-08
Refactoring readback support

ryan-summers:
This PR does the following:
* Exposes a `string_get()` API
* Refactors `recursive_iter()` to just iterate across the possible topic names (keys)
* Adds in an `iter` module that contains all iteration-related code.
* Merges `develop`
* Adds a `get_metadata()` API to determine maximum necessary topic + state buffer sizes

### DIFF
--- a/derive_miniconf/src/lib.rs
+++ b/derive_miniconf/src/lib.rs
@@ -22,11 +22,14 @@ impl TypeDefinition {
     /// # Note
     /// This is equivalent to adding:
     /// `where Self: DeserializeOwned` to the type definition.
-    pub fn add_deserialize_bound(&mut self) {
+    pub fn add_serde_bound(&mut self) {
         let where_clause = self.generics.make_where_clause();
         where_clause
             .predicates
             .push(parse_quote!(Self: miniconf::DeserializeOwned));
+        where_clause
+            .predicates
+            .push(parse_quote!(Self: miniconf::Serialize));
     }
 
     // Bound all generics of the type with `T: miniconf::DeserializeOwned + Miniconf`. This is necessary to
@@ -37,6 +40,7 @@ impl TypeDefinition {
                 type_param
                     .bounds
                     .push(parse_quote!(miniconf::DeserializeOwned));
+                type_param.bounds.push(parse_quote!(miniconf::Serialize));
                 type_param.bounds.push(parse_quote!(miniconf::Miniconf));
             }
         }
@@ -136,8 +140,8 @@ fn derive_struct(mut typedef: TypeDefinition, data: syn::DataStruct, atomic: boo
     // If this structure must be updated atomically, it is not valid to call Miniconf recursively
     // on its members.
     if atomic {
-        // Bound the Miniconf implementation on Self implementing DeserializeOwned.
-        typedef.add_deserialize_bound();
+        // Bound the Miniconf implementation on Self implementing DeserializeOwned + Serialize.
+        typedef.add_serde_bound();
 
         let name = typedef.name;
         let (impl_generics, ty_generics, where_clause) = typedef.generics.split_for_impl();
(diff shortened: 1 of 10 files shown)


derive_miniconf/src/lib.rs
```diff
@@ -174,14 +211,30 @@ fn derive_struct(mut typedef: TypeDefinition, data: syn::DataStruct, atomic: boo
...
-                topic.push_str(concat!("/", stringify!(#field_name))).unwrap();
-                if let Some(()) = self.#field_name.recursive_iter(&mut index[1..], topic, value) {
+                let postfix = if topic.len() != 0 {
+                    concat!("/", stringify!(#field_name))
+                } else {
+                    stringify!(#field_name)
+                };
+
+                if topic.push_str(postfix).is_err() {
+                    return None;
```
ryan-summers, line 221: We should probably log/error on this condition, since the user won't be getting all possible keys.

derive_miniconf/src/lib.rs
```diff
@@ -174,14 +211,30 @@ fn derive_struct(mut typedef: TypeDefinition, data: syn::DataStruct, atomic: boo
...
-                else {
-                    index[0] += 1;
-                    index[1..].iter_mut().for_each(|x| *x = 0);
+
+                // Strip off the previously prepended index, since we completed that element and need
+                // to instead check the next one.
+                while let Some(character) = topic.pop() {
+                    if character == '/' {
+                        break;
+                    }
```
irwineffect, line 233: We could backup the size of the topic before we push the new postfix, and then use `String::truncate()` to revert.

derive_miniconf/src/lib.rs
irwineffect: Agreed, not sure what we should do here. A panic seems too strong for a user error, I think what you have here is the best we can do, and then document what will happen if TS is too small in `into_iter<TS>()`

derive_miniconf/src/lib.rs
```diff
@@ -259,62 +320,34 @@ fn derive_enum(mut typedef: TypeDefinition, data: syn::DataEnum) -> TokenStream
...
+
+                miniconf::serde_json_core::to_slice(self, value).map_err(|_| miniconf::Error::SerializationFailed)
+            }
+
+            fn recursive_iter<const TS: usize>(&self, index: &mut [usize], topic: &mut miniconf::heapless::String<TS>) -> Option<()> {
+                if index.len() == 0 {
+                    // I don't expect this to happen...
+                    panic!("index stack too small");
+                    // return None;
+                }
```
irwineffect, line 338: Given that this is a user error, should we return None instead of panicking? If it's ok to panic here, should be ok to panic when TS is too small.

derive_miniconf/src/lib.rs
ryan-summers: I like that much more than the current approach - good idea, will implement

derive_miniconf/src/lib.rs
ryan-summers: I've refactored to add a `get_metadata()` function that can be used to retrieve the maximum topic size + recursive depth (length of state vector). This allows us to check the sizes before we even start iteration.

src/lib.rs
```diff
@@ -158,18 +176,47 @@ pub trait Miniconf {
...
+    /// * `TS` - The maximum number of bytes to encode a settings path into.
+    ///
+    /// # Args
+    /// * `state` - A state vector to record iteration state in.
     fn into_iter<'a, const TS: usize>(
         &'a self,
         state: &'a mut [usize],
-    ) -> iter::MiniconfIter<'a, Self, TS> {
-        iter::MiniconfIter {
+    ) -> Result<iter::MiniconfIter<'a, Self, TS>, IterError> {
```
irwineffect: I like this solution of checking buffer sizes at creation time of the iterator. Now that we guard the buffer sizes, I think it would make sense to panic if later code detects that we ran out of space, at that point it is an internal library error right? We wouldn't want to hide it by returning `None` earlier than expected. 

Also, may want to consider adding an `into_iter_unchecked()` version in case a user wants to perform the check only once?

ryan-summers: Technically the user can still hit these cases if they manually call `recurse_paths()`, but I'm fine with making it `unreachable!()`, since that's not the intended use case.

irwineffect (approved):

PR #66, opened 2021-12-10
Feature/settings republish

### LINKED ISSUE #10: Add setting re-publication (readback) support

ryan-summers:
This PR closes #10 by adding functionality to the MqttClient to automatically publish all available settings after 2 seconds of connection + last message received.

TODO:
* [x] Refactor static buffer sizes
* [x] Remove iterator state when iteration has completed
* [x] Add MQTT-republish integration test|

Note: https://github.com/quartiq/minimq/issues/74 might mean that a device will fail to publish settings on boot until that issue is fixed and incorporated into a new minimq patch.

### DIFF
--- a/src/mqtt_client/mqtt_client.rs
+++ b/src/mqtt_client/mqtt_client.rs
@@ -84,7 +84,7 @@ where
         // Check the settings topic length.
         let settings = Settings::default();
 
-        assert!(settings.get_metadata().max_topic_size > MAX_TOPIC_LENGTH);
+        assert!(settings.get_metadata().max_topic_size <= MAX_TOPIC_LENGTH);
 
         let mut mqtt = minimq::Minimq::new(broker, client_id, stack, clock.clone())?;
 
@@ -102,7 +102,7 @@ where
                 &connection_topic,
                 "0".as_bytes(),
                 QoS::AtMostOnce,
-                Retain::NotRetained,
+                Retain::Retained,
                 &[],
             )
             .unwrap();
@@ -125,6 +125,7 @@ where
     fn handle_republish(&mut self) {
         if let Some(timeout) = &self.rx_timeout {
             if self.clock.try_now().unwrap() > *timeout {
+                log::info!("Republish duration lapsed. Publishing settings");
                 self.iteration_state.replace([0; 16]);
             }
         }
@@ -149,7 +150,7 @@ where
                 let len = self.settings.get(&topic, &mut data).unwrap();
 
                 let mut prefixed_topic: String<{ MAX_TOPIC_LENGTH }> = String::new();
-                write!(&mut prefixed_topic, "{}/{}", &self.prefix, &topic).unwrap();
+                write!(&mut prefixed_topic, "{}/{}", &self.settings_prefix, &topic).unwrap();
 
                 // Note(unwrap): This should not fail because `can_publish()` was checked before
                 // attempting this publish.
@@ -159,7 +160,7 @@ where
                         &prefixed_topic,
                         &data[..len],
                         QoS::AtMostOnce,
-                        Retain::Retained,
+                        Retain::NotRetained,
                         &[],
                     )
                     .unwrap();
--- a/src/mqtt_client/mqtt_client.rs
+++ b/src/mqtt_client/mqtt_client.rs
@@ -29,25 +29,38 @@ use minimq::{embedded_time, QoS, Retain};
 
 use core::fmt::Write;
 
+// The maximum topic length of any settings path.
+const MAX_TOPIC_LENGTH: usize = 64;
+
+// The maximum prefix length of the settings topic.
+const MAX_PREFIX_LENGTH: usize = 64;
+
+// The keepalive interval to use for MQTT in seconds.
+const KEEPALIVE_INTERVAL_SECONDS: u16 = 60;
+
+// The delay after not receiving messages after initial connection that settings will be
+// republished.
+const REPUBLISH_DELAY_SECS: u32 = 2;
+
 /// MQTT settings interface.
-pub struct MqttClient<Settings, Stack, Clock, const MESSAGE_SIZE: usize, const MESSAGE_COUNT: usize>
+pub struct MqttClient<Settings, Stack, Clock, const MESSAGE_SIZE: usize>
 where
     Settings: Miniconf + Default,
     Stack: TcpClientStack,
     Clock: embedded_time::Clock,
 {
-    mqtt: minimq::Minimq<Stack, Clock, MESSAGE_SIZE, MESSAGE_COUNT>,
+    mqtt: minimq::Minimq<Stack, Clock, MESSAGE_SIZE, 1>,
     clock: Clock,
     settings: Settings,
     subscribed: bool,
-    settings_prefix: String<64>,
-    prefix: String<64>,
+    settings_prefix: String<MAX_PREFIX_LENGTH>,
+    prefix: String<MAX_PREFIX_LENGTH>,
     iteration_state: Option<[usize; 16]>,
     rx_timeout: Option<Instant<Clock>>,
 }
 
-impl<Settings, Stack, Clock, const MESSAGE_SIZE: usize, const MESSAGE_COUNT: usize>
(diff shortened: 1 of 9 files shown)


src/mqtt_client/mqtt_client.rs
```diff
@@ -83,18 +107,81 @@ where
...
+
+            iteration_exhausted = true;
+        }
+
+        // If iteration has been exhausted, clear the iterator state, as we are now done
+        // republishing.
+        if iteration_exhausted {
+            self.iteration_state.take();
+        }
+    }
```
jordens, line 183: I don't think you need `iteration_exhausted`. The only paths out of the loop is either return or completing. The latter then must `take()` in all cases.

And shouldn't we more cleanly store an `Option<Iterator> and not its opaque internal state? I would create and own the iterator state in the iterator struct.

src/mqtt_client/mqtt_client.rs
```diff
@@ -115,7 +203,8 @@ where
             self.subscribed = true;
```
jordens, line 203: This is now big enough so that it should not be a collection of flags (`subscribed`, `Option<Timeout>`, `Option<Iterator>`, `iteration_exhausted`) but a state machine. Let's do this comprehensively.

src/mqtt_client/mqtt_client.rs
```diff
@@ -166,6 +263,10 @@ where
 
             let response = MqttMessage::new(properties, default_response_topic, &message);
 
+            if rx_timeout.is_some() {
```
jordens, line 266: We should check whether the message is retained (assuming we can). If the message is not retained, don't advance the timeout.

jordens (changes requested):

src/mqtt_client/mqtt_client.rs
ryan-summers: The iterator struct cannot exist outside of this function because it has to borrow a reference to `settings`. However, `settings` is mutably borrowed in `update()`, so we need to release any immutable references to it inbetween. That's why the iterator is intended to go out of scope when this function exits, but the iteration information is retained in `state` so we can resume from where we left off.

src/mqtt_client/mqtt_client.rs
ryan-summers: Yeah, I was noticing this as well. Will refactor

src/mqtt_client/mqtt_client.rs
ryan-summers: Right now, I don't think minimq exposes the flags of the received message, just the topic, data, and properties, so we don't know if they are retained or not.

src/mqtt_client/mqtt_client.rs
jordens: Ah. Interesting problem. I would still make the state more opaque though and ensure type coherence between saving and resuming. Then maybe more generically a trait `ResumableIterator<T>` with `save(self) -> Self::State` and `resume(Self::State) -> Self`?

src/mqtt_client/mqtt_client.rs
jordens: Yeah. There is a race condition (retained timeout racing new settings coming in). But from the potential consequences it looks benign. The only general issue I see is that active settings updates may effectively DOS republication. I think that can only be solved by a dirty flag for each path (received a retained message) and an timeout since connection instead. Fine for now.

src/mqtt_client/mqtt_client.rs
ryan-summers: I would personally prefer if we pulled discussion of that out of this PR and into a refactor of the `Miniconf` trait if we feel it to be necessary. Right now, the iterator state is not exposed to the end user, so there's no need to typestate it. It will always be assocaited with the `MqttClient::Settings` type and is internal only.

jordens: Ok. But I still have the suspicion that the following are common problems where it is somewhat likely that someone else has solved them: (a) suspending and resuming an iterator (b) iterating over the fields of a struct (there seems to be [`field_names`](https://github.com/TedDriggs/field_names)). Especially all this introspection stuff sounds related.
I'd rather evaluate those other solutions early than later having/wanting to migrate.

ryan-summers: Stringifying the names is trivial in a proc-macro. That is _not_ what we have done, and is not where the complexity of this solution comes from.

What we have done is allowed recursive iteration of nested structures by mapping unique indices to each field (in a recursive manner). 

James and I evaluated a number of various crates and found that none did _quite_ what we wanted. Similarly, this isn't necessarily an issue of just pausing/resuming iteration that can be solved easily because this has to fight with Rust's ownership system on the thing we're iterating. AFAICT there is no existing `SuspendableIterator` trait.

While we could make one, I don't think it actually either helps us significantly

ryan-summers: If you want to look at the change set related to _just_ iteration over the settings structure, the PR is https://github.com/quartiq/miniconf/pull/62

src/mqtt_client/mqtt_client.rs
```diff
@@ -112,73 +168,92 @@ where
...
+            .unwrap();
+    }
+
+    fn handle_subscription(&mut self) {
+        log::info!("MQTT connected, subscribing to settings");
 
-            iteration_exhausted = true;
+        // Note(unwrap): We construct a string with two more characters than the prefix
+        // structure, so we are guaranteed to have space for storage.
+        let mut settings_topic: String<{ MAX_TOPIC_LENGTH + 2 }> =
```
jordens, line 228: I don't think it makes sense to fine grain this. Let's just assert that all topics are MAX_TOPIC_LENGTH.

```suggestion
        let mut settings_topic: String<{ MAX_TOPIC_LENGTH }> =
```

src/mqtt_client/mqtt_client.rs
```diff
@@ -112,73 +168,92 @@ where
...
+            .into_iter::<{ MAX_TOPIC_LENGTH }>(&mut self.state.context_mut().republish_state)
+            .unwrap()
+        {
+            let mut data = [0; MESSAGE_SIZE];
+
+            // Note(unwrap): We know this topic exists already because we just got it from the
+            // iterator.
+            let len = self.settings.get(&topic, &mut data).unwrap();
 
+            let mut prefixed_topic: String<{ 2 * MAX_TOPIC_LENGTH + 1 }> = String::new();
```
jordens, line 194: Increase it to 128 max topic length. Since we are juggling one topic at a time in all cases, the stack usage will be flat.

```suggestion
            let mut prefixed_topic: String<{ MAX_TOPIC_LENGTH }> = String::new();
```

jordens (changes requested):

src/mqtt_client/mqtt_client.rs
ryan-summers: Looking into `field_names` looks like it'd be somewhat helpful, but it'd also require us to rely on the user using `#[derive(FieldNames, Miniconf)]` because we cannot invoke proc macros from other proc macros. If they forgot the derive, It's likely they'd get extremely opaque error messages about not being able to find T::FIELDS

jordens: Ack. I haven't seen the details of evaluation and I can't find it in the PR.
On a quick search I also was unable to find something like a `SuspendableIterator` trait but maybe you agree that this is potentially a worthwhile generalization.
I do think generalizing these would help even if it only forces the analysis of the consistency and soundness of this design.
For now as an MVP this is ok. But my feeling is that it will either break or migrate in the mid-term.

jordens (approved):

src/mqtt_client/mqtt_client.rs
ryan-summers: I've opened an issue into investigating already-created suspendable iteration as https://github.com/quartiq/miniconf/issues/69

PR #82, opened 2022-02-14
Implementing fix to ignore incoming republish data

### LINKED ISSUE #71: Settings republish events register as updated settings
When miniconf republishes settings, it registers the message that it has published as inbound change requests. This causes it to generate settings update events for all settings.

ryan-summers:
This PR fixes #71 by adding correlation data to the republished settings so they can be quickly ignored.

TODO:
- https://github.com/quartiq/minimq/pull/81 needs to be merged and 0.5.2 of minimq needs to be released


jordens (changes requested):
Nice. There are two helpful clippy lints, other than that this looks good.

jordens (approved):

PR #89, opened 2022-05-12
Renaming iterator functions to prevent name conflicts

### LINKED ISSUE #87: `Miniconf::iter()` method conflicts with `core::slice::iter()`
The `iter()` method can cause confusion when being called on array objects.

Because of [Rust's expression resolution ordering](https://doc.rust-lang.org/reference/expressions/method-call-expr.html), an array does not natively implement the `iter()` method. Instead, that function is implemented on slices. Because of this, calling `iter()` on an array will find `Miniconf::iter()` instead of trying to convert the array to a slice and then looking up functions.

In this case:
1. Rust fails to find the `iter()` method on the [array object](https://doc.rust-lang.org/std/primitive.array.html) because it indeed has no `iter()` function.
2. Rust looks at trait impls for `iter()` on the array object. In our case, it finds `iter()` because of `trait Miniconf` and uses that

However, this may be undesirable because the user might be trying to coerce the array object into a slice to call the slice iterator method. Because of the resolution ordering, the Miniconf `iter()` method is selected instead.

The workaround for this defect is to first explicitly convert the array object into a slice as follows:
```rust
let mut array = [0f32; 5];
// Finds Miniconf::iter();
array.iter()
// Finds core::slice::iter
array[..].iter();
```

We likely want to rename the `iter()` method to prevent this confusion in the future.

ryan-summers:
This PR fixes #87 by renaming the `iter()` methods to `iter_settings()` to avoid conflicting with the resolution lookup ordering of `core::slice::iter()`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Changed
+* **breaking** The Miniconf trait for iteration was renamed from `unchecked_iter()` and `iter()` to
+  `unchecked_iter_settings()` and `iter_settings()` respectively to avoid issues with slice iteration
+  name conflicts. See [#87](https://github.com/quartiq/miniconf/issues/87)
+
 ## [0.4.0] - 2022-05-11
 
 ### Added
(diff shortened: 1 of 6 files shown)


jordens (approved):
good!

PR #92, opened 2022-07-07
py: remove whitespace on dump

jordens:
* this matches the serde-json-core behavior
* also saves a bit of space
* note that without these whitespaces the generated json is not a subset
  of yaml anymore. this will not matter jere as the target must parse it as json, not yaml.


ryan-summers (approved):
LGTM. Nice change to cut down on size since we don't care about spaces.

PR #96, opened 2022-10-13
Support heapless primitives

nkrackow:
It would be nice to have heapless string support for miniconf out of the box. 
I tested this working but you can probably fumble it into the macro as well? Sorry, I'm not an expert on macros and I didn't want to spend more time on this now.  Maybe this is easy for you @ryan-summers? :)


src/lib.rs
```diff
@@ -407,3 +408,63 @@ impl_single!(f64);
 
 impl_single!(usize);
 impl_single!(bool);
+
+impl<const N: usize> Miniconf for String<N> {
```
ryan-summers, line 412: This indeed looks identical to the implementation for `impl_single! with the caveat that it has the const-generic surrounding it. There's likely a way we can wrap these together. I'm happy to take over if you don't want to worry about it.

nkrackow: Cool, I spent a little while looking into macros but I decided it's too long of an endeavor for me right now if you maybe already know how to.

jordens (approved):
Nice!

jordens:
What about `Vec`?

nkrackow:
Recursive macros, sure.. ;)
Thanks!

jordens:
Note that the macro should not be needed at all. It could be a plain generic impl. The only reason the macro is needed is to exclude arrays from the generic impl (no negative trait bounds or negative impl yet).

jordens:
Maybe that's an indication that having the custom impl of array is a suboptimal solution.

ryan-summers:
> The only reason the macro is needed is to exclude arrays from the generic impl

Correct. Is there a way we can exclude arrays in a generic manner from the blanket impl? I'm not so sure yet.

src/array.rs
```diff
@@ -2,126 +2,143 @@ use super::{Error, Miniconf, MiniconfMetadata};
...
+    };
+}
+
+impl<T: Miniconf, const N: usize> Miniconf for [T; N] {
+    impl_array!();
+}
 
-        None
-    }
+impl<T: Miniconf, const N: usize> Miniconf for heapless::Vec<T, N> {
```
jordens, line 142: Hmm. I had thought `Vec` should be like `String`, i.e. atomic. Otherwise there is no way to ever change the length of a `Vec`.

jordens: Or better tag individual members with `#[miniconf(atomic)]` in the enclosing `#[derive(Miniconf)]` struct (this is roughly how other crates like `thiserror` control the `derive`).

jordens: But all this is a bigger redesign. I'd propose landing this PR with `String` an `Vec` being terminal and then spawning an issue for the redesign.

ryan-summers: Honestly, `Vec` is a whole different beast because of how versatile it is. I would recommend we limit this to just `String` and figure out how to handle `Vec` later.

jordens: It would be a one-liner to add support for `Vec` in the same way as for `String`.

ryan-summers: Yes, but I don't think that's how we would actually want it implemented. In my opinion, a Vec is much like an Option, where you have a variable-length list of configurable settings, not something that should be mutated directly via external means.

jordens: IMO it's the opposite. `Vec` should be atomic and thus trivially support variable length data. `array` can be non-terminal.
(That is before the potential refactor along the lines described above.)

But OK. Let's have this PR only cover `String`.

PR #98, opened 2022-10-19
Feature/unique discovery

### LINKED ISSUE #97: Python Discovery may discover the same device twice
When using automatic device discovery, the same device may be discovered twice as a result of the `alive` indication message.

This generally occurs if a device is rebooted shortly before discovery is required.

ryan-summers:
This PR fixes #97 by updating the `discover` python utility to return only unique device prefixes. This prevents the same device being discovered twice interfering with automated single-device instantiation.

### DIFF
--- a/py/miniconf-mqtt/miniconf/__init__.py
+++ b/py/miniconf-mqtt/miniconf/__init__.py
@@ -47,4 +47,4 @@ async def discover(
 
     await asyncio.sleep(discovery_timeout)
 
-    return discovered_devices
+    return list(set(discovered_devices))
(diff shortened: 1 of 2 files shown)


py/miniconf-mqtt/miniconf/__init__.py
```diff
@@ -47,4 +47,4 @@ def handle_message(_client, topic, payload, _qos, _properties):
 
     await asyncio.sleep(discovery_timeout)
 
-    return discovered_devices
+    return list(set(discovered_devices))
```
jordens, line 50: I would change the API and return a set (also start with a set instead of a list in `discovered_devices = set() ... discovered_devices.add(topic[:-len(suffix)]) ...`)

jordens (changes requested):

jordens (approved):

PR #100, opened 2022-10-28
Feature/trait rework

### LINKED ISSUE #99: Redesign atomic/terminal elements
As discussed in https://github.com/quartiq/miniconf/pull/96#discussion_r999242502 we could consider:

* Maybe we need to split our single trait into `Miniconf` and `MiniconfAtomic` for the following:
* Instead of a macro [`impl_single`](https://github.com/quartiq/miniconf/blob/90eea767543e2dfd787e2f923604b9d2d5830b7d/src/lib.rs#L330), blanket `impl MiniconfAtomic for T where T: Serialize + Deserialize`
* That extends MiniconfAtomic support to everything that is deserializable/serializable but makes them terminal/atomic elements.
* That then also covers `array` and give it atomic support
* Would also cover every struct that is `Serialize + Deserialize`
* Keep the current custom (inner) `Miniconf` (explicitly or derive) support for `array`, `Option` etc
* Add (inner) `Miniconf` support for `Vec` and others that are [`IndexMut`](https://doc.rust-lang.org/core/ops/trait.IndexMut.html#) + `IntoIterator` (?)
* Consider (inner) `Miniconf` for things that are container-like: tuples (macro impl up to 32 tuple size) or hashmaps.
* Add support for tagging specific struct members with `miniconf(defer)` (modulo bikesheding for a better name: `defer`, `inner`, `non-atomic`) to the `derive(Miniconf)` macro. This is along the lines how other crates like e.g. `thiserror` achieve fine grained control over derive macros (`error(transparent)` etc).
* That would allow deferring to the (inner) `Miniconf` impl within the member and make the member non-atomic.
* Otherwise and by default a struct member is exposed through its `MiniconfAtomic`.
* Maybe allow exposing both the outer and inner `Miniconf` explicitly in (the rare) cases where this is not an ambiguity problem.
* (Re)Consider supporting `derive(Miniconf)` for tuples, enums.
* This gets rid of the ambiguity of whether `array`/`String`/`Vec` are `Atomic` or not, i.e. whether they use inner or outer miniconf.

Not completely sure whether this fully maps onto the trait architecture of rust though and whether there are no conflicting trait impls or incompatibilities.
Not sure whether the traits `Miniconf` and `MiniconfAtomic` can/should/need to be disjoint for this.

```rust

#[derive(Serialize, Deserialize, Miniconf)]
struct Settings {
    a: [0u8; 2],  // "a = [0, 1]"
    #[miniconf(defer)]
    b: [0u8; 8],  // "b/1 = 1"
}
```

ryan-summers:
This PR fixes #99 

Notable changes:
* A new `#[miniconf(defer)]` tag has replaced `#[derive(MiniconfAtomic)]`. This tells miniconf to recurse through the object for settings. Any item not explicitly marked as `defer` will be trated as an atomic element.
* New `MiniconfArray` type has been added
   * The `MiniconfArray` is used in leiu of a `[T; N]` to defer down to miniconf traits when using deference.
      * A default implementation is provided for `[T; N]` that defers down to serde primitives instead of Miniconf traits.

* Default Miniconf implementations have been removed for basic types. Instead, Miniconf relies on `Serialize + DeserializeOwned` on these types, which allows integrating foreign types into `Miniconf` structures without having to implement `Miniconf` on them
* Trait bounding for generics has been simplified to relax the requirements of Serde traits to only those that directly are called.

### DIFF
--- a/derive_miniconf/src/field.rs
+++ b/derive_miniconf/src/field.rs
@@ -1,0 +1,48 @@
+
+use syn::parse_quote;
+use super::{TypeDefinition, attributes::{MiniconfAttribute, AttributeParser}};
+
+pub struct StructField {
+    pub field: syn::Field,
+    pub deferred: bool,
+}
+
+impl StructField {
+    pub fn new(field: syn::Field) -> Self {
+        let deferred = field.attrs.iter().filter(|attr| attr.path.is_ident("miniconf")).map(|attr| AttributeParser::new(attr.tokens.clone()).parse()).any(|x| x == MiniconfAttribute::Defer);
+
+        Self {
+            deferred,
+            field,
+        }
+    }
+
+    pub fn bound_generics(&self, typedef: &mut TypeDefinition) {
+        // Check our type. Path-like types may need to be bound.
+        let path = match &self.field.ty {
+            syn::Type::Path(syn::TypePath { path, ..}) => path,
+            _ => return,
+        };
+
+        // Generics will have an ident only as the type. Grab it.
+        let ident = match path.get_ident() {
+            Some(ident) => ident,
+            _ => return,
+        };
+
+        // If we got here, we may have a generic parameter. Check the ident against all of the
+        // generic params on the type.
+        for generic in &mut typedef.generics.params {
+            if let syn::GenericParam::Type(type_param) = generic {
+                if type_param.ident == *ident {
+                    if self.deferred {
+                        type_param.bounds.push(parse_quote!(miniconf::Miniconf));
+                    } else {
+                        type_param.bounds.push(parse_quote!(miniconf::Serialize));
+                        type_param.bounds.push(parse_quote!(miniconf::DeserializeOwned));
+                    }
+                }
+            }
+        }
+    }
+}
--- a/derive_miniconf/src/field.rs
+++ b/derive_miniconf/src/field.rs
@@ -1,6 +1,8 @@
-
+use super::{
+    attributes::{AttributeParser, MiniconfAttribute},
+    TypeDefinition,
+};
 use syn::parse_quote;
-use super::{TypeDefinition, attributes::{MiniconfAttribute, AttributeParser}};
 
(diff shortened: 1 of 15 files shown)


derive_miniconf/src/field.rs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+impl StructField {
+    pub fn new(field: syn::Field) -> Self {
+        let deferred = field
+            .attrs
+            .iter()
+            .filter(|attr| attr.path.is_ident("miniconf"))
+            .map(|attr| AttributeParser::new(attr.tokens.clone()).parse())
+            .any(|x| x == MiniconfAttribute::Defer);
```
ryan-summers, line 19: It seems like any `Array` type should also be deferred by-default. It's relatively rare that a user would want to set the entire array in a single invocation.

src/option.rs
```diff
@@ -1,12 +1,70 @@
...
+
+impl<T: Clone + Miniconf> Clone for OptionalSetting<T> {
+    fn clone(&self) -> Self {
+        Self(self.0.clone())
+    }
+}
+
+impl<T: Copy + Miniconf> Copy for OptionalSetting<T> {}
+
+impl<T: Miniconf> Miniconf for OptionalSetting<T> {
```
ryan-summers, line 61: Should we call this `DeferredOption` to be in-line with `DeferredArray`?

jordens (approved):

jordens (changes requested):
Correction. I meant to "request changes"...

jordens (changes requested):
Very nice. Let me know whether I should update https://github.com/quartiq/miniconf/pull/102 or whether we wait with that.

ryan-summers:
@jordens Lets land #102 right after this. No need to introduce more thrash immediately. I like those naming cleanups though, and we're about to break the API anyways, so it's a good chance to clean it up.

jordens (approved):

PR #101, opened 2022-10-28
Cleaning up python client

### LINKED ISSUE #57: Add timeout to Python API
Add a request timeout via the Python API. If the device requested is not present, a command may hang indefinitely. Instead, we should raise some sort of timeout.

### LINKED ISSUE #84: Analyze retain-by-default Python client behavior
In many cases, it seems counter intuitive that settings configured via python are retained-by-default.

For test scripts, this means every call to `command()` has to have `retained=False`. Instead, it may make sense to have the CLI operate in retained-by-default, but make the API exposed over python only operate as retained-when-specified.

### LINKED ISSUE #94: Refactor python client to utilize UUID CorrelationData
Replace the response topic string with just a static response topic.

Additionally, update the `CorrelationData` property to use the UUID instead of an int.

ryan-summers:
This PR fixes #94 by updating the python client to store UUID as the correlation data and use a static response topic.

This PR fixes #84 by updating the python API to be no-retain by default.

This PR fixes #57 by adding a timeout to the Python API.

### DIFF
--- a/py/miniconf-mqtt/miniconf/__main__.py
+++ b/py/miniconf-mqtt/miniconf/__main__.py
@@ -58,8 +58,8 @@ def main():
         assert len(devices) == 1, \
             f'Multiple miniconf devices found ({devices}). Please specify a more specific --prefix'
 
-        logging.info('Automatically using detected device prefix: %s', list(devices)[0])
         prefix = list(devices)[0]
+        logging.info('Automatically using detected device prefix: %s', prefix)
 
     async def configure_settings():
         interface = await Miniconf.create(prefix, args.broker)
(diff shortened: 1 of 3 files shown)


py/miniconf-mqtt/miniconf/__main__.py
```diff
@@ -58,8 +58,8 @@ def main():
         assert len(devices) == 1, \
             f'Multiple miniconf devices found ({devices}). Please specify a more specific --prefix'
 
-        logging.info('Automatically using detected device prefix: %s', devices[0])
-        prefix = devices[0]
+        prefix = list(devices)[0]
```
jordens, line 61:
```suggestion
        prefix = devices.pop()
```

jordens (changes requested):
One little thing. Otherwise LGTM

jordens (approved):

Issue #104, opened 2022-11-01
be generic over the path format and separator


jordens:
~No real need to enforce a `Peekable<Split<char>>`.
Maybe https://docs.rs/itertools/latest/itertools/trait.PeekingNext.html is enough.~

* [x] `get_path()` and `set_path()` are now (with #105) generic over the path format and separator.
* [ ] `next_path` and `metadata()` need redesign to make them agnostic.

ryan-summers:
There are a few places where a separator is employed:
1. During derivation to construct setting lookup paths
2. During `Miniconf::get()` and `Miniconf::set()` default implementations.
3. In default `miniconf::Array` and `[T; N]` implementations

It seems like the macro could (quite easily) expose a new attribute to specify the separator:
```rust
#[derive(Miniconf)]
#[miniconf(separator=':')]
struct Settings {
   // ...
}
```

During code-gen, we could implement an overriden `get()` and `set()` function as well as updated path generation function. I'm not sure how we could fix up the default implementation for `array` - perhaps the separator could be passed into the `recurse_paths` function.

jordens:
Or maybe `next_path` can be restructured to return an iterator over path parts. Then the user (or `iter_paths()`) can join each path.

PR #105, opened 2022-11-01
rework (make get/set_path generic Peekable, associated fns for metadata and next_path)

### LINKED ISSUE #88: Add mutable iteration support
When iterating across the settings array, the settings structure is immutably borrowed. This means that settings cannot be modified during iteration.

It would be useful to allow mutable iteration of settings.

There are some implications here. If we allow mutably iterating, we will need to specifically address iterating over `Option`s. If the option were replaced during iteration, it could have unintended consequences on structure iteration.

jordens:
* a small step towards #104
* Changes the path assembly logic to postfix the separator. This is a bit easier in the logic.
* makes `metadata()` an associated fn
* homogenizes the API and e.g. makes the mqtt client and the unittests use `set()` where appropriate
* also makes the metadata semantically correct (depth N if there are N elements in the path)
* make `next_path` and related API associated fns
* Move the iterator state into the iterator
* fixes #88 by iterating all paths regardless of `Option` nulls.
* adds `Error::PathAbsent` to mark runtime absence in the case of `Option::None`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -332,7 +332,7 @@ pub trait Miniconf {
         &'a self,
         state: &'a mut [usize],
     ) -> Result<iter::MiniconfIter<'a, Self, TS>, IterError> {
-        let meta = self.metadata();
+        let meta = Self::metadata();
 
         if TS < meta.max_length {
             return Err(IterError::InsufficientTopicLength);
@@ -391,5 +391,5 @@ pub trait Miniconf {
     ) -> bool;
 
     /// Get metadata about the structure.
-    fn metadata(&self) -> Metadata;
+    fn metadata() -> Metadata;
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -277,6 +277,7 @@ pub struct Metadata {
     pub max_depth: usize,
 }
 
+/// Helper trait for `core::iter::Peekable`.
 pub trait Peekable: core::iter::Iterator {
     fn peek(&mut self) -> core::option::Option<&Self::Item>;
 }
@@ -292,7 +293,7 @@ pub trait Miniconf {
     /// Update an element by path.
     ///
     /// # Args
-    /// * `path` - The path to the element.
+    /// * `path` - The path to the element. With '/' as separator.
     /// * `data` - The serialized data making up the content.
     ///
     /// # Returns
@@ -304,7 +305,7 @@ pub trait Miniconf {
     /// Retrieve a serialized value by path.
     ///
     /// # Args
-    /// * `path` - The path to the element.
+    /// * `path` - The path to the element. With '/' as separator.
     /// * `data` - The buffer to serialize the data into.
     ///
     /// # Returns
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -119,8 +119,7 @@
 //!
 //! let settings = Settings::default();
 //!
-//!let mut state = [0; 8];
-//! for topic in settings.iter_paths::<128>(&mut state).unwrap() {
+//! for topic in Settings::iter_paths::<8, 128>().unwrap() {
 //!     println!("Discovered topic: `{:?}`", topic);
 //! }
 //! ```
@@ -318,31 +317,22 @@ pub trait Miniconf {
     ///
     /// This is a depth-first walk.
     ///
-    /// # Note
-    /// To start the iteration from the first path,
-    /// the state vector should be initialized with zeros.
-    /// The state vector can be used to resume iteration from a previous point in time.
-    ///
     /// # Template Arguments
-    /// * `TS` - The maximum number of bytes to encode a path into.
-    ///
-    /// # Args
-    /// * `state` - A state vector to record iteration state in.
-    fn iter_paths<'a, const TS: usize>(
-        &'a self,
-        state: &'a mut [usize],
-    ) -> Result<iter::MiniconfIter<'a, Self, TS>, IterError> {
+    /// * `L`  - The maximum depth of the path, i.e. number of separators plus 1.
+    /// * `TS` - The maximum length of the path in bytes.
+    fn iter_paths<const L: usize, const TS: usize>(
+    ) -> Result<iter::MiniconfIter<Self, L, TS>, IterError> {
         let meta = Self::metadata();
 
         if TS < meta.max_length {
             return Err(IterError::InsufficientTopicLength);
         }
 
-        if state.len() < meta.max_depth {
+        if L < meta.max_depth {
             return Err(IterError::InsufficientStateDepth);
         }
 
-        Ok(self.unchecked_iter_paths(state))
+        Ok(Self::unchecked_iter_paths())
     }
 
     /// Create an iterator of all possible paths.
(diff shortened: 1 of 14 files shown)


src/lib.rs
```diff
@@ -277,30 +277,41 @@ pub struct Metadata {
...
 pub trait Miniconf {
     /// Update an element by path.
     ///
     /// # Args
-    /// * `path` - The path to the element.
+    /// * `path` - The path to the element. With '/' as separator.
     /// * `data` - The serialized data making up the content.
     ///
     /// # Returns
     /// The result of the configuration operation.
```
ryan-summers, line 300: Should update docs to indicate it returns the number of bytes used for deserialization

src/lib.rs
```diff
@@ -277,30 +277,41 @@ pub struct Metadata {
...
+    }
+}
+
 /// Derive-able trait for structures that can be mutated using serialized paths and values.
 pub trait Miniconf {
     /// Update an element by path.
     ///
     /// # Args
-    /// * `path` - The path to the element.
+    /// * `path` - The path to the element. With '/' as separator.
```
ryan-summers, line 296:
```suggestion
    /// * `path` - The path to the element with '/' as the separator.
```

src/lib.rs
```diff
@@ -277,30 +277,41 @@ pub struct Metadata {
...
-        self.set_path(path.split('/').peekable(), data)
+    fn set(&mut self, path: &str, data: &[u8]) -> Result<usize, Error> {
+        self.set_path(&mut path.split('/').peekable(), data)
     }
 
     /// Retrieve a serialized value by path.
     ///
     /// # Args
-    /// * `path` - The path to the element.
+    /// * `path` - The path to the element. With '/' as separator.
```
ryan-summers, line 308:
```suggestion
    /// * `path` - The path to the element with '/' as the separator.
```

src/lib.rs
```diff
@@ -277,30 +277,41 @@ pub struct Metadata {
...
     /// * `data` - The buffer to serialize the data into.
     ///
     /// # Returns
     /// The number of bytes used in the `data` buffer for serialization.
     fn get(&self, path: &str, data: &mut [u8]) -> Result<usize, Error> {
-        self.get_path(path.split('/').peekable(), data)
+        self.get_path(&mut path.split('/').peekable(), data)
```
ryan-summers, line 314: Why pass a mutable reference instead of handing ownership for the iterator through the chain? There's no reason we need to retain ownership here.

src/option.rs
```diff
@@ -95,51 +96,64 @@ impl<T: Miniconf> Miniconf for Option<T> {
...
+        }
+
         if path_parts.peek().is_some() {
             return Err(Error::PathTooLong);
         }
 
-        *self = serde_json_core::from_slice(value)?.0;
-        Ok(())
+        let (value, len) = serde_json_core::from_slice(value)?;
+        *self = Some(value);
```
ryan-summers, line 118:
```suggestion
        *self = value;
```

The deserialized `value` is an `Option` type, and thus may be `null`, in which case it is `None`.

src/option.rs
```diff
@@ -95,51 +96,64 @@ impl<T: Miniconf> Miniconf for Option<T> {
...
+            return Err(Error::PathNotFound);
+        }
+
         if path_parts.peek().is_some() {
             return Err(Error::PathTooLong);
         }
 
-        serde_json_core::to_slice(self, value).map_err(|_| Error::SerializationFailed)
+        // Note(unwrap): checked above
+        serde_json_core::to_slice(self.as_ref().unwrap(), value)
```
ryan-summers, line 136:
```suggestion
        let data = self.as_ref().ok_or(Error::PathNotFound)?;
        serde_json_core::to_slice(data, value)
```

I think we can get rid of the unwrap and note entirely.

tests/option.rs
```diff
@@ -62,19 +62,49 @@ fn option_test_normal_option() {
...
+    s.data = Some(0);
+    s.set("data", b"7").unwrap();
+    assert_eq!(s.data, Some(7));
+
+    let mut state = [0; 10];
+    let mut iterator = s.iter_paths::<128>(&mut state).unwrap();
+    assert_eq!(iterator.next(), Some("data".into()));
+    assert!(iterator.next().is_none());
+
+    assert!(s.set("data", b"null").is_err());
```
ryan-summers, line 109: I think it could be desirable for the user to be able to do this for one of the option types. This would allow e.g. disabling some function via settings

ryan-summers (changes requested):

src/mqtt_client/mod.rs
```diff
@@ -177,7 +177,7 @@ where
         let mut settings_prefix: String<MAX_TOPIC_LENGTH> = String::from(prefix);
         settings_prefix.push_str("/settings").unwrap();
 
-        assert!(settings_prefix.len() + 1 + settings.metadata().max_length <= MAX_TOPIC_LENGTH);
+        assert!(settings_prefix.len() + 1 + Settings::metadata().max_length <= MAX_TOPIC_LENGTH);
 
         Ok(Self {
             mqtt,
```
ryan-summers, line 183: Path iteration on the settings tree needs to be updated to handle non-existent `get()` responses from `Option`s

src/option.rs
jordens: For `Option` we only support setting `Some`. Otherwise the user (through Miniconf) could set a `Some` to a `None` exactly once and then never `get()` or `set()` it.

src/option.rs
jordens: But it is already checked before in L120. And we want that for error consistency (not found before too long).

tests/option.rs
jordens: That's already possible by using a plain non-deferred `core::Option`.
Otherwise a deferred `core::Option` would be no different from the non-deferred one.

src/lib.rs
jordens: Within `get()` no.
But for `get_path()` the iterator may be expensive to create and we don't need to own it. Since there is no need to own it and ownership being more restrictive than a mutable borrow while not being faster, the mutable borrow is preferrable IMO.

jordens:
@ryan-summers this rework PR escalated a bit. I'll stop before doing more significant changes.

src/option.rs
jordens: OTOH might actually be better to do it that way and add a specific `PathAbsent` error for runtime absence.

src/option.rs
jordens: This is also a bit of a problem in `serde_json{,_core}`. For `Option<()>` both variants are JSON `null`.

src/option.rs
ryan-summers: My point being we can avoid an unwrap. Feel free to put the `as_ref().ok_or()` earlier in the function where it is right now

ryan-summers (approved):
LGTM. Nice sets of changes :)

src/option.rs
jordens: Agreed. The unwrap is gone now.

Issue #103, opened 2022-11-01
be generic over the serde implementation


jordens:
* [ ] it doesn't need to be JSON for the values
* [x] the path items for array indices could be parsed via some other means, e.g. `str::parse::<usize>()` (#105)

ryan-summers:
I believe it should be relatively easy to make a `Miniconf` impl generic over a `Serializer` or `Deserializer` that we call into directly.
We could easily also provide defaults for serde_json_core as well if desired.

The indices should definitely be `FromStr` parseable.

jordens:
or even just `set_path()` generic over `Deserializer` and `get_path` generic over `Serializer`.

ryan-summers:
Hmm... One annoyance is that `serde_json_core` does not expose the `Serializer` and `Deserializer` via the public API. That being said, I don't see why we couldn't make a PR to add that support. The user would be responsible for constructing the serializer/deserializer from a buffer, but that's no big ask.

jordens:
Ack. We can still expose that easier serde-json-core API that does that for the user in the same way we expose `set()` that splits the path for the user before calling `set_path()`.

PR #108, opened 2022-11-02
symmetric error

jordens:
also use implicit from

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -202,7 +202,7 @@ pub use heapless;
 
 /// Errors that can occur when using the `Miniconf` API.
 #[non_exhaustive]
-#[derive(Debug, PartialEq)]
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -272,6 +272,12 @@ impl From<serde_json_core::de::Error> for Error {
     }
 }
 
+impl From<serde_json_core::ser::Error> for Error {
+    fn from(err: serde_json_core::ser::Error) -> Error {
+        Error::Serialization(err)
```
ryan-summers, line 277: Thanks! This is something I noticed recently too - we definitely don't want to just drop the serializer error

jordens: Yes. It was fine up to now since there is only exactly one serializer error which is just buffer full as the comment says. But that;as certainly not to be relied on and problematic for #103

ryan-summers (approved):

PR #109, opened 2022-11-02
next_path() errors and path count

jordens:


src/iter.rs
```diff
@@ -24,10 +35,16 @@ impl<M: Miniconf + ?Sized, const L: usize, const TS: usize> Iterator for Minicon
     fn next(&mut self) -> Option<Self::Item> {
         let mut path = Self::Item::new();
 
-        if M::next_path(&mut self.state, &mut path) {
+        if M::next_path(&mut self.state, &mut path).unwrap() {
```
ryan-summers, line 38: Having this unwrap sucks. It's fine for our usecase, but introduces format bloat into applications that otherwise may not have it.

Some options:
1. Return a `Result<String, IterErr>` as the `Item`
2. match on the error and generate internal panic (doesn't seem much different than just unwrap IMO)
3. Somehow handle and log the error instead of failing
4. Leave it as is and indicate that this is a design limitation

I think (1) would be my preference? I'm not entirely sure. What are your thoughts?

src/iter.rs
```diff
@@ -7,13 +7,24 @@ use heapless::String;
 pub struct MiniconfIter<M: ?Sized, const L: usize, const TS: usize> {
     marker: PhantomData<M>,
     state: [usize; L],
+    count: Option<usize>,
```
ryan-summers, line 10: Mind adding some docs for what these members are supposed to be? It took me a minute to grok what `count` is supposed to be. We should also document what `state` is

src/lib.rs
```diff
@@ -338,29 +342,35 @@ pub trait Miniconf {
...
     ///
+    /// # Args
+    /// * count: Optional iterator length
+    ///
     /// # Template Arguments
     /// * `L`  - The maximum depth of the path, i.e. number of separators plus 1.
     /// * `TS` - The maximum length of the path in bytes.
-    fn unchecked_iter_paths<const L: usize, const TS: usize>() -> iter::MiniconfIter<Self, L, TS> {
-        iter::MiniconfIter::default()
+    fn unchecked_iter_paths<const L: usize, const TS: usize>(
```
ryan-summers, line 370: Also needs a changelog entry, as this is technically (another) breaking API change

ryan-summers (changes requested):

src/iter.rs
jordens: This PR replaces all the `unreachable!()` panics and their format bloat with this single one. As you say, the only real alternative is a fallible iterator that would be used with `let mut it = m.iter_paths(); while let Some(ele) = it.next()? { ... }`.
But it only moves having to deal with the error up one level. Because of that, I think re-evaluating the mqtt client with its numerous unwraps() is a lower hanging fruit.

src/iter.rs
jordens: Ack

src/iter.rs
ryan-summers: I'm not sure that `panic!()` and `unreachable!()` have the same panic bloat as an `unwrap()`. If the user specifies `panic` as an `abort`, you don't get format bloat, but rather just panic data bloat (metadata about where the panic occurred) in RODATA.

jordens: Given the more annoying DX of a fallible iterator, I'd lean towards (in a future PR when we re-evaluate the MQTT client) add that in parallel to the existing panicing one.

jordens: > If the user specifies panic as an abort, you don't get format bloat (I think?),

Right. Same in both cases.

https://doc.rust-lang.org/src/core/result.rs.html#1817-1827

ryan-summers (approved):

PR #114, opened 2022-11-04
docs update

jordens:


README.md
```diff
@@ -1,28 +1,149 @@
...
+Miniconf is designed to be protocol-agnostic. Any means that can receive key-value input from
+some external source can be used to modify values by path.
+
+## Limitations
+Deferred (non-atomic) access to inner elements of some types is not yet supported. This includes:
+* Complex enums (other than [core::option::Option])
+* Tuple structs (other than [Option], [Array])
+
+## Features
+* `mqtt-client` Enabled the MQTT client feature. See the example in [MqttClient].
```
ryan-summers, line 149: While these links are valid for `rust-doc`, they are not valid markdown. As such, most of the links to rust code in this doc are broken:

image

src/iter.rs
```diff
@@ -16,8 +16,8 @@ pub struct MiniconfIter<M: ?Sized, const L: usize, const TS: usize> {
 
     /// The remaining length of the iterator.
     ///
-    /// It is used to provide `Iterator::size_hint()`
-    /// and may be None to indicate unknown length.
+    /// It is used to provide an exact and trusted `Iterator::size_hint()`.
```
ryan-summers, line 19: I'm not sure what "trusted" is supposed to mean. I don't really think it's needed to say that it's correct...

README.md
jordens: I know. But I think that's a minor issue, less critical than having redundant and divergent docs in README vs the library docstrings. I'm happy to have a slightly ugly README if in turn the content is much better and in sync with the docs.

ryan-summers: Fair enough. We can address it later if it becomes annoying

src/iter.rs
jordens: It is important information as this goes beyond the size_hint promises.
It would be `TrustedLen` if that was stable.

ryan-summers (approved):

PR #116, opened 2022-11-05
reorganize derive macro

jordens:
* refactor the derive macro a bit, move out the arms into functions, remove `TypeDefinition`, cancel some common code
* derive common traits for Array and Option
* support zero length arrays and structs

### DIFF
--- a/derive_miniconf/src/lib.rs
+++ b/derive_miniconf/src/lib.rs
@@ -153,10 +153,9 @@ fn metadata_arm((i, struct_field): (usize, &StructField)) -> proc_macro2::TokenS
 /// Derive the Miniconf trait for structs.
 ///
 /// # Args
-/// * `input` - The derive input token stream.
 /// * `data` - The data associated with the struct definition.
-/// * `atomic` - specified true if the data must be updated atomically. If false, data must be
-///   set at a terminal node.
+/// * `generics` - The generics of the definition. Sufficient bounds will be added here.
+/// * `ident` - The identifier to derive the impl for.
 ///
 /// # Returns
 /// A token stream of the generated code.
--- a/derive_miniconf/src/lib.rs
+++ b/derive_miniconf/src/lib.rs
@@ -187,6 +187,7 @@ fn derive_struct(mut input: DeriveInput) -> TokenStream {
                 value: &[u8]
             ) -> Result<usize, miniconf::Error> {
                 let field = path_parts.next().ok_or(miniconf::Error::PathTooShort)?;
+                #[allow(unreachable_code)]
                 let peek = path_parts.peek().is_some();
 
                 match field {
@@ -201,6 +202,7 @@ fn derive_struct(mut input: DeriveInput) -> TokenStream {
                 value: &mut [u8]
             ) -> Result<usize, miniconf::Error> {
                 let field = path_parts.next().ok_or(miniconf::Error::PathTooShort)?;
+                #[allow(unreachable_code)]
                 let peek = path_parts.peek().is_some();
 
                 match field {
@@ -213,9 +215,9 @@ fn derive_struct(mut input: DeriveInput) -> TokenStream {
                 state: &mut [usize],
                 path: &mut miniconf::heapless::String<TS>
             ) -> Result<bool, miniconf::IterError> {
+                #[allow(unreachable_code)]
+                let original_length = path.len();
                 loop {
-                    let original_length = path.len();
-
                     match *state.first().ok_or(miniconf::IterError::PathDepth)? {
                         #(#next_path_arms ,)*
                         _ => return Ok(false),
@@ -240,14 +242,17 @@ fn derive_struct(mut input: DeriveInput) -> TokenStream {
                 let mut meta = miniconf::Metadata::default();
 
                 for index in 0.. {
-                    let item_meta = match index {
+                    let item_meta: miniconf::Metadata = match index {
                         #(#metadata_arms ,)*
                         _ => break,
                     };
 
-                    meta.max_length = meta.max_length.max(item_meta.max_length);
-                    meta.max_depth = meta.max_depth.max(item_meta.max_depth);
-                    meta.count += item_meta.count;
+                    #[allow(unreachable_code)]
+                    {
+                        meta.max_length = meta.max_length.max(item_meta.max_length);
+                        meta.max_depth = meta.max_depth.max(item_meta.max_depth);
+                        meta.count += item_meta.count;
+                    }
                 }
 
                 meta
--- a/derive_miniconf/src/lib.rs
+++ b/derive_miniconf/src/lib.rs
@@ -7,18 +7,6 @@ mod field;
 
 use field::StructField;
 
-/// Represents a type definition with associated generics.
-pub(crate) struct TypeDefinition {
-    pub generics: syn::Generics,
-    pub name: syn::Ident,
-}
-
-impl TypeDefinition {
-    pub fn new(generics: syn::Generics, name: syn::Ident) -> Self {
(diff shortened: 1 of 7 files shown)


derive_miniconf/src/lib.rs
```diff
@@ -49,162 +37,157 @@ impl TypeDefinition {
...
+///
+/// # Args
+/// * `input` - The derive input token stream.
+/// * `data` - The data associated with the struct definition.
+/// * `atomic` - specified true if the data must be updated atomically. If false, data must be
+///   set at a terminal node.
+///
+/// # Returns
+/// A token stream of the generated code.
+fn derive_struct(mut input: DeriveInput) -> TokenStream {
```
ryan-summers, line 159: Why take in the `input` and then desugar it into `syn::Data::Struct(data)` again after we already do this at the top level?

derive_miniconf/src/lib.rs
```diff
@@ -49,162 +37,157 @@ impl TypeDefinition {
...
+}
 
-    let metadata_arms = fields.iter().enumerate().map(|(i, f)| {
-        let field_type = &f.field.ty;
-        let field_name = &f.field.ident;
-        if f.deferred {
-            quote! {
-                #i => {
-                    let mut meta = <#field_type>::metadata();
+fn metadata_arm((i, f): (usize, &StructField)) -> proc_macro2::TokenStream {
```
ryan-summers, line 117: Should we make these functions members of `StructField` instead?

ryan-summers, line 117: Please rename `f` to `field` or `struct_field` throughout. `f` isn't a very helpful variable name

derive_miniconf/src/lib.rs
```diff
@@ -49,162 +37,157 @@ impl TypeDefinition {
...
+    if f.deferred {
+        quote! {
+            stringify!(#match_name) => {
+                self.#match_name.get_path(path_parts, value)
+            }
+        }
+    } else {
+        quote! {
+            stringify!(#match_name) => {
+                if peek {
```
ryan-summers, line 57: While this is efficient, it's not great from a code maintainability perspective. The `peek` variable is defined pretty far away from the rest of this code, so it may not be clear where it's supposed to be coming from.

IMO we should likely just rely on Rust being smart enough to optimize the redundant `peek()` calls. I'm sure it can do it without issue if the field isn't mutating.

derive_miniconf/src/lib.rs
```diff
@@ -218,45 +201,61 @@ fn derive_struct(mut typedef: TypeDefinition, data: syn::DataStruct) -> TokenStr
...
+            ) -> Result<bool, miniconf::IterError> {
+                let original_length = path.len();
+                loop {
+                    match *state.first().ok_or(miniconf::IterError::PathDepth)? {
+                        #(#next_path_arms ,)*
+                        _ => return Ok(false),
+                    };
+
+                    // Note(unreachable) Without any deferred fields, every arm above returns
+                    #[allow(unreachable_code)]
```
ryan-summers, line 224: Can (should?) we check for at least 1 deferred field before calling into this? Not sure if it's worth it though.

src/array.rs
```diff
@@ -16,7 +16,8 @@ use super::{Error, IterError, Metadata, Miniconf, Peekable};
 use core::fmt::Write;
 
 /// An array that exposes each element through their [`Miniconf`](trait.Miniconf.html) implementation.
-pub struct Array<T, const N: usize>(pub [T; N]);
+#[derive(Clone, Copy, PartialEq, Eq, Debug, PartialOrd, Ord, Hash)]
+pub struct Array<T, const N: usize>([T; N]);
```
ryan-summers, line 20: Will this break if you try to supply a `T` that isn't e.g. `PartialEq`? I don't think we want to restrict what types the user can supply, as they may be out of the end-user's control.

src/option.rs
```diff
@@ -19,7 +19,20 @@
...
+    Copy,
+    Default,
+    PartialEq,
+    Eq,
+    Debug,
+    PartialOrd,
+    Ord,
+    Hash,
+    serde::Serialize,
+    serde::Deserialize,
```
ryan-summers, line 33: Similar to above, what if the `T` doesn't impl `Serialize` or `Deserialize`? It's not required to based on our impl, since we defer downward into the `Miniconf` impl

ryan-summers (changes requested):

derive_miniconf/src/lib.rs
jordens: That would not help with your concern https://github.com/quartiq/miniconf/pull/116/files#r1015249137

jordens: It has always been `f`.

derive_miniconf/src/lib.rs
jordens: In `derive()` we're only checking for `data` variant to dispatch to the right function.
Here we also need other fields (other than `data`).

derive_miniconf/src/lib.rs
jordens: Since it's just dead code even in the case where the compiler notices it (it does) but doesn't optimize it away, I think this is minor.

src/option.rs
jordens: It does the right thing.

derive_miniconf/src/lib.rs
jordens: The same argument (readability) holds for `value`, `path_parts`, `self` and the rest of the control context.

src/array.rs
ryan-summers: Rust playground agrees! Wonderful news. Thanks for clarifying.

src/option.rs
ryan-summers: Why derive `Serialize` and `Deserialize` here, but not on `Array`?

derive_miniconf/src/lib.rs
ryan-summers: Fair point

src/option.rs
jordens: Can't. Same as default.

derive_miniconf/src/lib.rs
ryan-summers: Scratch this. I was thinking `structfield` was a type we defined, but it's pulled in from `syn`

jordens: But otherwise yes. Ich think that's nicer.

jordens: Ack

src/option.rs
ryan-summers: Ah right, there's the annoying issue of only implementing things through [T; 32] that has held around... https://docs.rs/serde/1.0.147/serde/ser/index.html#implementations-of-serialize-provided-by-serde

I think this could be done via conditional impls if we wanted, but we can deal with this later if it ever is an issue.

derive_miniconf/src/lib.rs
jordens: I don't see a clear alternative.

ryan-summers: You could just destructure in `derive` and immediately pass in `generics` and `name` as input args to `derive_struct`

jordens: I fear that would be more code than the match

ryan-summers: I don't think so? I think it's just two additional args for the function for `input.ident` and `input.generics`?

ryan-summers (approved):

ryan-summers (approved):

PR #121, opened 2022-11-14
add from/into docs

jordens:


src/option.rs
```diff
@@ -1,24 +1,26 @@
...
-/// An `Option` that exposes its value through their [`Miniconf`](trait.Miniconf.html) implementation.
+use core::ops::{Deref, DerefMut};
+
+/// An `Option` that exposes its value through their [`Miniconf`] implementation.
+///
+/// # Design
+///
+/// Miniconf supports optional values in two forms. The first for is the [`miniconf::Option`](Option)
+/// type. If the `Option` is `None`, the part of the namespace does not exist at run-time.
+/// It will not be iterated over and cannot be `get()` or `set()` using the [`Miniconf`] API.
```
ryan-summers, line 10: Isn't this doc no longer true? i.e. you will always iterate over `None` topic trees regardless, but they generate a unique error during `set()` and `get()`

ryan-summers (changes requested):

ryan-summers (approved):

PR #128, opened 2022-12-10
option: actually return PathAbsent

jordens:


ryan-summers (approved):
LGTM. Thanks for the fix :)

PR #130, opened 2022-12-10
rj/itoa

jordens:
- array: use itoa, remove fmt/write usage
- mqtt_client: remove write/fmt usage

### DIFF
--- a/src/mqtt_client/mod.rs
+++ b/src/mqtt_client/mod.rs
@@ -220,7 +220,9 @@ where
             };
 
             let mut prefixed_topic: String<MAX_TOPIC_LENGTH> = String::new();
-            write!(&mut prefixed_topic, "{}/{}", &self.settings_prefix, &topic).unwrap();
+            prefixed_topic.push_str(&self.settings_prefix).unwrap();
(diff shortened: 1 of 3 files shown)


src/mqtt_client/mod.rs
```diff
@@ -207,29 +205,30 @@ where
...
 
+        let prefix = String::from(prefix);
+
         // Configure a will so that we can indicate whether or not we are connected.
-        let mut connection_topic: String<MAX_TOPIC_LENGTH> = String::from(prefix);
+        let mut connection_topic = prefix.clone();
         connection_topic.push_str("/alive").unwrap();
         mqtt.client()
             .set_will(
                 &connection_topic,
```
ryan-summers, line 215: Can this just be `&prefix`? The borrow should end before ownership of `prefix` moves I believe

ryan-summers (approved):
One minor nit

src/mqtt_client/mod.rs
jordens: It needs the `/alive` appended.

PR #133, opened 2022-12-14
Updating response codes to use user properties

### LINKED ISSUE #131: Utilize `UserProperty` instead of `Response` for status codes
Instead of having double serialization in the `Response` data type, just append the code to the list of properties and parse it on the host-side software.

See https://github.com/quartiq/miniconf/pull/124#discussion_r1047442122 for more info

ryan-summers:
This PR fixes #131 by leveraging `UserProperty` for response codes to avoid double-serialization issues.

This also lessens stack pressure by removing the need to allocate large buffers in some instances.


jordens (approved):
Very nice

PR #141, opened 2023-05-09
array: AsRef, AsMut, IntoIterator

jordens:


tests/arrays.rs
```diff
@@ -74,6 +74,28 @@ fn simple_array_indexing() {
...
+        a: miniconf::Array<miniconf::Array<Inner, 2>, 2>,
+    }
+
+    let mut s = S::default();
+
+    for _i in s.a.into_iter().flatten() {}
+
+    for _i in s.a.iter().flatten() {}
+
+    for _i in s.a.iter_mut().flatten() {}
```
ryan-summers, line 96: Can we make these actually iterate across something to verify iteration works? Seems odd to me to do nothing here. Maybe even just an `iter().collect()` and compare to a known vec?

ryan-summers (approved):

PR #143, opened 2023-05-12
fix subscription topic

### LINKED ISSUE #142: Interference/problems with miniconf and other mqtt paths
Using the latest master branch with eg. Stabilizer dual-iir (aka using vanilla Stabilizer master branch, just changing the miniconf dep to the github master and minimq to 0.6.2) leads to the following issues:

Miniconf seems to be confused with messages that are not intended for it. I get the following debug prints:
```
INFO - Starting
INFO - EUI48: 80-34-28-1f-59-65
INFO - setup() complete
WARN - Network link DOWN
WARN - Network link UP
INFO - MQTT connected, subscribing to settings
INFO - Unknown Miniconf command: /alive
INFO - Unknown Miniconf command: /telemetry
```

If I publish onto a miniconf setting I get `WARN - Failed to build response message`.

The miniconf settings are applied at the device and I also see the mqtt messages like `/alive` on the host.

jordens:
close #142

### DIFF
--- a/src/mqtt_client/mod.rs
+++ b/src/mqtt_client/mod.rs
@@ -318,10 +318,8 @@ where
     fn handle_subscription(&mut self) {
         log::info!("MQTT connected, subscribing to settings");
 
-        // Note(unwrap): We construct a string with two more characters than the prefix
-        // structure, so we are guaranteed to have space for storage.
         let mut settings_topic = self.prefix.clone();
-        settings_topic.push_str("/#").unwrap();
+        settings_topic.push_str("/settings/#").unwrap();
 
         let topic_filter = TopicFilter::new(&settings_topic)
             .options(SubscriptionOptions::default().ignore_local_messages());


src/mqtt_client/mod.rs
```diff
@@ -318,10 +318,8 @@ where
     fn handle_subscription(&mut self) {
         log::info!("MQTT connected, subscribing to settings");
 
-        // Note(unwrap): We construct a string with two more characters than the prefix
-        // structure, so we are guaranteed to have space for storage.
         let mut settings_topic = self.prefix.clone();
-        settings_topic.push_str("/#").unwrap();
+        settings_topic.push_str("/settings/#").unwrap();
```
ryan-summers, line 322: I'm confused. The subscription options should have prevented the alive and telemetry messages

ryan-summers: Yeah, we expect to get the settings prefix and subscribe at the upper level because we parse the commands get/set that occur below the settings tree

https://github.com/quartiq/miniconf/blob/8be449dcd9be7c838720c21223dac0841595dd26/src/mqtt_client/mod.rs#L97

jordens: Not telemetry. That's a different client. But the alive, yes. Maybe.

jordens: Ah right.

PR #145, opened 2023-06-22
Bumping client, releasing Miniconf

ryan-summers:
This PR releases miniconf as 0.7

I'll follow up with a python release to pip as well following this now that things have stabilized a bit.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,5 +34,10 @@ std-embedded-nal = "0.1"
 tokio = { version = "1.9", features = ["rt-multi-thread", "time", "macros"] }
 std-embedded-time = "0.1"
 
+[patch.crates-io.std-embedded-nal]
+version = "0.1"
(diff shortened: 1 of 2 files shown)


jordens (approved):

Cargo.toml
```diff
@@ -34,5 +34,10 @@ std-embedded-nal = "0.1"
 tokio = { version = "1.9", features = ["rt-multi-thread", "time", "macros"] }
 std-embedded-time = "0.1"
 
+[patch.crates-io.std-embedded-nal]
```
jordens, line 37: Nice work patching this as dev-dep and releasing nonetheless! ;)

ryan-summers: Dev dependencies aren't used by end-users, only locally for testing etc. so I have no issue releasing with dev-dep patches IMO. Very low risk

jordens: Yes. Just pointing that out.

PR #146, opened 2023-06-22
only reply if response topic set

### LINKED ISSUE #142: Interference/problems with miniconf and other mqtt paths

jordens:
* [x] only try to build response if ResponseTopic is set
* [x] should subscribe to `/settings/#` and `/list` explicitly, not to `/#` (which captures telemetry, alive, and maybe even response topics, log etc as well)

closes #142

### DIFF
--- a/src/mqtt_client/mod.rs
+++ b/src/mqtt_client/mod.rs
@@ -232,11 +232,11 @@ where
     }
 
     fn handle_listing(&mut self) {
-        let Some(ref mut iter) = &mut self.listing_state else {
+        let Some(iter) = &mut self.listing_state else {
             return;
         };
 
-        let Some(ref props) = self.properties_cache else {
+        let Some(props) = &self.properties_cache else {
             return
         };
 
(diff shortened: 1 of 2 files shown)


src/mqtt_client/mod.rs
```diff
@@ -96,21 +96,20 @@ enum Command<'a> {
...
-                if value.is_empty() {
-                    Ok(Command::Get { path })
-                } else {
-                    Ok(Command::Set { path, value })
+
+        if path == "list" {
+            Ok(Command::List)
+        } else {
+            match path.split_once('/') {
+                Some(("settings", path)) => {
```
ryan-summers, line 104: Does this properly check that `path` is not None?

jordens: In this arm `path` is always `&str`.

ryan-summers (approved):

src/mqtt_client/mod.rs
```diff
@@ -442,20 +446,10 @@ where
         F: FnMut(&str, &mut Settings, &Settings) -> Result<(), E>,
         E: AsRef<str>,
     {
-        let Self {
```
ryan-summers, line 445: Huh, I wonder what changed that made this no longer necessary. I'm glad we can get rid of it in any case!

ryan-summers (changes requested):
Can you please add relevant CHANGELOG entries for the fact that we fixed an issue about subscriptions and that we now only respond if there's a response topic?

src/mqtt_client/mod.rs
jordens: Yeah. This became unnecessary a while ago. It now accurately captures individual items in structs (and not the entire struct). Also https://doc.rust-lang.org/1.33.0/book/ch18-03-pattern-syntax.html#legacy-patterns-ref-and-ref-mut

ryan-summers (approved):

PR #148, opened 2023-07-23
be generic over serde mechanism and path hierarchy separator

### LINKED ISSUE #103: be generic over the serde implementation

### LINKED ISSUE #104: be generic over the path format and separator

jordens:
Close #104 
Close #103 

Tested in `fls` where it also makes the code 1.2 kB smaller.

Downside is that we loose serde error source information.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-* The `Miniconf` trait is now generic over the `Deserializer`/`Serializer`. It
+* [breaking] The `Miniconf` trait is now generic over the `Deserializer`/`Serializer`. It
   doesn't enforce `serde-json-core` or `u8` buffers anymore.
-* `MiniconfIter` takes the path hierarchy separator and passes it on to
+* [breaking] `MiniconfIter` takes the path hierarchy separator and passes it on to
   `Miniconf::next_path`.
-* The `Miniconf` trait has been stripped of the provided functions that depended
+* [breaking] The `Miniconf` trait has been stripped of the provided functions that depended
   on the `serde`-backend and path hierarchy separator. Those have been
   moved into a super trait `SerDe<S>` that is generic over a specification marker
   struct `S`. `SerDe<JsonCoreSlash>` has been implemented for all `Miniconf`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+* The `Miniconf` trait is now generic over the `Deserializer`/`Serializer`. It
+  doesn't enforce `serde-json-core` or `u8` buffers anymore.
+* `MiniconfIter` takes the path hierarchy separator and passes it on to
+  `Miniconf::next_path`.
+* The `Miniconf` trait has been stripped of the provided functions that depended
+  on the `serde`-backend and path hierarchy separator. Those have been
+  moved into a super trait `SerDe<S>` that is generic over a specification marker
+  struct `S`. `SerDe<JsonCoreSlash>` has been implemented for all `Miniconf`
+  to provide the previously existing functionality.
+* The only required change for most downstream crates to adapt to the above is to
+  make sure the `SerDe` trait is in scope (`use miniconf::SerDe`).
+
 ## [0.7.1] (https://github.com/quartiq/miniconf/compare/v0.7.0...v0.7.1)
 
 ### Fixed
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ rust-version = "1.65.0"
 
 [dependencies]
 miniconf_derive = { path = "miniconf_derive" , version = "0.6" }
-serde-json-core = { git = "https://github.com/quartiq/serde-json-core.git", branch = "rj/ser-de-pub" }
+serde-json-core = { git = "https://github.com/rust-embedded-community/serde-json-core.git" }
 serde = { version = "1.0.120", features = ["derive"], default-features = false }
 log = "0.4"
 heapless = { version = "0.7", features = ["serde"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ rust-version = "1.65.0"
 
 [dependencies]
 miniconf_derive = { path = "miniconf_derive" , version = "0.6" }
-serde-json-core = {path = "../serde-json-core"} # "0.5.0"
+serde-json-core = { git = "https://github.com/quartiq/serde-json-core.git", branch = "rj/ser-de-pub" }
 serde = { version = "1.0.120", features = ["derive"], default-features = false }
 log = "0.4"
 heapless = { version = "0.7", features = ["serde"] }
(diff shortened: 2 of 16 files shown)


miniconf_derive/src/lib.rs
```diff
@@ -58,7 +58,7 @@ fn get_path_arm(struct_field: &StructField) -> proc_macro2::TokenStream {
                 if peek {
                     Err(miniconf::Error::PathTooLong)
                 } else {
-                    Ok(miniconf::serde_json_core::to_slice(&self.#match_name, value)?)
+                    serde::ser::Serialize::serialize(&self.#match_name, ser).map_err(|_| miniconf::Error::Serialization)
```
ryan-summers, line 61: We need to fully qualify the `serde` path to something like `miniconf::serde` - `serde` may not be a direct dependent in the user crate.

miniconf_derive/src/lib.rs
```diff
@@ -80,9 +80,8 @@ fn set_path_arm(struct_field: &StructField) -> proc_macro2::TokenStream {
                 if peek {
                     Err(miniconf::Error::PathTooLong)
                 } else {
-                    let (value, len) = miniconf::serde_json_core::from_slice(value)?;
-                    self.#match_name = value;
-                    Ok(len)
+                    self.#match_name = serde::de::Deserialize::deserialize(de).map_err(|_| miniconf::Error::Deserialization)?;
```
ryan-summers, line 83: Fully-qualify `serde`

src/lib.rs
```diff
@@ -54,13 +54,13 @@ pub enum Error {
 
     /// The value provided for configuration could not be deserialized into the proper type.
     ///
-    /// Check that the serialized data is valid JSON and of the correct type.
-    Deserialization(serde_json_core::de::Error),
+    /// Check that the serialized data is valid and of the correct type.
+    Deserialization,
```
ryan-summers, line 58: Is there a way we could have the de::Error and ser::Error part of the SerDe trait and make the Error enum generic over that?

I've encountered trouble when I discarded error information because it makes it really hard to backtrace where things went wrong, and in this case, I think it traces back all the way over the MQTT python interface to the user, so we should maintain as much information about the error as possible to aid the user in fixing it.

src/lib.rs
```diff
@@ -140,127 +117,142 @@ impl<I: core::iter::Iterator> Peekable for core::iter::Peekable<I> {
...
+        separator: char,
+    ) -> Result<bool, IterError>;
+
+    /// Get metadata about the paths in the namespace.
+    fn metadata() -> Metadata;
+}
+
+pub trait SerDe<S>: Miniconf {
+    /// The path hierarchy separator.
+    const SEPARATOR: char;
```
ryan-summers, line 175: Mind adding a note that this separator type is used by the path iterators? I couldn't figure out why it was a member of the trait for a long time, and we should document the usage of it for docs.rs

src/option.rs
```diff
@@ -134,22 +135,21 @@ impl<T: crate::Serialize + crate::DeserializeOwned> Miniconf for core::option::O
             return Err(Error::PathAbsent);
         }
 
-        let (value, len) = serde_json_core::from_slice(value)?;
-        *self = Some(value);
-        Ok(len)
+        *self = Some(serde::Deserialize::deserialize(de).map_err(|_| Error::Deserialization)?);
```
ryan-summers, line 138: Hmm, it's occurring to me that we no longer handle a "data remains after deserialization" as an error. So technically, something like: `{"data": 5}blahblahblah` would still deserialize a `Map` properly, wouldn't it?

src/option.rs
```diff
@@ -162,12 +162,13 @@ impl<T: crate::Serialize + crate::DeserializeOwned> Miniconf for core::option::O
     fn next_path<const TS: usize>(
         state: &mut [usize],
         path: &mut heapless::String<TS>,
+        separator: char,
     ) -> Result<bool, IterError> {
         if *state.first().ok_or(IterError::PathDepth)? == 0 {
             state[0] += 1;
 
             // Remove trailing slash added by a deferring container (array or struct).
```
ryan-summers, line 170:
```suggestion
            // Remove trailing separator added by a deferring container (array or struct).
```

CHANGELOG.md
```diff
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
...
 
+## Unreleased
+
+### Changed
+
+* The `Miniconf` trait is now generic over the `Deserializer`/`Serializer`. It
+  doesn't enforce `serde-json-core` or `u8` buffers anymore.
+* `MiniconfIter` takes the path hierarchy separator and passes it on to
+  `Miniconf::next_path`.
+* The `Miniconf` trait has been stripped of the provided functions that depended
```
ryan-summers, line 15:
```suggestion
* [breaking] The `Miniconf` trait has been stripped of the provided functions that depended
```

Cargo.toml
```diff
@@ -15,7 +15,7 @@ rust-version = "1.65.0"
 
 [dependencies]
 miniconf_derive = { path = "miniconf_derive" , version = "0.6" }
-serde-json-core = "0.5.0"
+serde-json-core = { git = "https://github.com/quartiq/serde-json-core.git", branch = "rj/ser-de-pub" }
```
ryan-summers, line 18: Can this get updated now? Maybe serde-json-core can cut a release

README.md
```diff
@@ -136,9 +137,11 @@ atomic access to their respective inner element(s), [Array] and
 into the inner element(s) through their respective inner [Miniconf] implementations.
 
 ## Formats
-The path hierarchy separator is the slash `/`.
 
-Values are serialized into and deserialized from JSON.
+Miniconf is generic over the `serde` backend/payload format and the path hierarchy separator
+(as long as the path can be split by it unambiguously).
+
+Currently support for `/` as the path hierarchy separator and JSON (`serde_json_core`) is implemented.
```
ryan-summers, line 144: I'd really like to see docs improved here. Maybe:
1. Show the user what an example of the default `json + slash` syntax looks like
2. Describe how the user could implement their own wire + path format.

ryan-summers (changes requested):
This looks quite nice. Some minor improvement requests

Cargo.toml
jordens: #149 does that.
Let's do the release prep in a second step.

README.md
jordens: The example above comprehensively shows what that syntax looks like.
I think the best example on how to implement a `SerDe<Spec>` (that'll be maintained and tested) is the one we have. So we should just point at that.

src/lib.rs
jordens: Yes. As mentioned that's a downside.
I tried and it didn't look very manageable. Might give it a nother spin.

src/option.rs
jordens: That's handled in `Serde::set()` (the `de.end()`).

src/lib.rs
ryan-summers: If it looks like a lot of work, let's spawn an issue out to handle it later then. Definitely something that needs fixing

README.md
ryan-summers: I think we need to add at least _some_ documentation to the `SerDe` trait to indicate what the intent is. Otherwise, I doubt any one would ever use the "custom serializer + path" syntax

jordens: Added docs to `SerDe`

PR #149, opened 2023-07-25
refactor path

jordens:
* Removes `Peekable`
* Any `Write` works a path for `next_path`, path additions don't need to be rolled back
* All `state` altering is now done in `MiniconfIter`
* Paths now start with the separator if not empty. This is consistent (paths never end in in the separator and each hierarchy level is prefixed with the separator) and allows getting rid of corner cases and really simplifies the code.
* [x] On top of/after #148

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,16 +9,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 * The `Miniconf` trait is now generic over the `Deserializer`/`Serializer`. It
-  doesn't enforce `serde-json-core` or `u8` buffers anymore.
-* `MiniconfIter` takes the path hierarchy separator and passes it on to
-  `Miniconf::next_path`.
+  doesn't enforce `serde-json-core` or `u8` buffers or `/` as the path hierarchy
+  separator anymore.
 * The `Miniconf` trait has been stripped of the provided functions that depended
   on the `serde`-backend and path hierarchy separator. Those have been
   moved into a super trait `SerDe<S>` that is generic over a specification marker
   struct `S`. `SerDe<JsonCoreSlash>` has been implemented for all `Miniconf`
   to provide the previously existing functionality.
+* `MiniconfIter` takes the path hierarchy separator from `SerDe` and passes it on to
+  `Miniconf::next_path`.
 * The only required change for most downstream crates to adapt to the above is to
   make sure the `SerDe` trait is in scope (`use miniconf::SerDe`).
+* Paths now start with the path separator (unless they are empty).
+* The path iterator does not need to be `Peekable` anymore.
 
 ## [0.7.1] (https://github.com/quartiq/miniconf/compare/v0.7.0...v0.7.1)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,16 +8,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-* The `Miniconf` trait is now generic over the `Deserializer`/`Serializer`. It
+* [breaking] The `Miniconf` trait is now generic over the `Deserializer`/`Serializer`. It
   doesn't enforce `serde-json-core` or `u8` buffers or `/` as the path hierarchy
   separator anymore.
-* The `Miniconf` trait has been stripped of the provided functions that depended
+* [breaking] `MiniconfIter` takes the path hierarchy separator from `SerDe` and passes it on to
+  `Miniconf::next_path`.
+* [breaking] The `Miniconf` trait has been stripped of the provided functions that depended
   on the `serde`-backend and path hierarchy separator. Those have been
   moved into a super trait `SerDe<S>` that is generic over a specification marker
   struct `S`. `SerDe<JsonCoreSlash>` has been implemented for all `Miniconf`
   to provide the previously existing functionality.
-* `MiniconfIter` takes the path hierarchy separator from `SerDe` and passes it on to
-  `Miniconf::next_path`.
 * The only required change for most downstream crates to adapt to the above is to
   make sure the `SerDe` trait is in scope (`use miniconf::SerDe`).
 * Paths now start with the path separator (unless they are empty).
--- a/miniconf_derive/src/lib.rs
+++ b/miniconf_derive/src/lib.rs
@@ -119,13 +119,8 @@ fn metadata_arm((i, struct_field): (usize, &StructField)) -> proc_macro2::TokenS
         quote! {
             #i => {
                 let mut meta = <#field_type>::metadata();
-
-                // Unconditionally account for separator since we add it
-                // even if elements that are deferred to (`Options`)
-                // may have no further hierarchy to add and remove the separator again.
                 meta.max_length += 1 + stringify!(#field_name).len();
                 meta.max_depth += 1;
-
                 meta
             }
         }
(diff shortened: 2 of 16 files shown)


CHANGELOG.md
```diff
@@ -9,16 +9,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 * The `Miniconf` trait has been stripped of the provided functions that depended
   on the `serde`-backend and path hierarchy separator. Those have been
   moved into a super trait `SerDe<S>` that is generic over a specification marker
   struct `S`. `SerDe<JsonCoreSlash>` has been implemented for all `Miniconf`
   to provide the previously existing functionality.
+* `MiniconfIter` takes the path hierarchy separator from `SerDe` and passes it on to
+  `Miniconf::next_path`.
 * The only required change for most downstream crates to adapt to the above is to
   make sure the `SerDe` trait is in scope (`use miniconf::SerDe`).
+* Paths now start with the path separator (unless they are empty).
```
ryan-summers, line 23: Probably a good idea :)

miniconf_derive/src/lib.rs
```diff
@@ -124,25 +119,18 @@ fn metadata_arm((i, struct_field): (usize, &StructField)) -> proc_macro2::TokenS
         quote! {
             #i => {
                 let mut meta = <#field_type>::metadata();
-
-                // Unconditionally account for separator since we add it
-                // even if elements that are deferred to (`Options`)
-                // may have no further hierarchy to add and remove the separator again.
```
ryan-summers, line 130: This doc is important and seems to me to be still acurate. Why'd we remove it?

miniconf_derive/src/lib.rs
```diff
@@ -124,25 +119,18 @@ fn metadata_arm((i, struct_field): (usize, &StructField)) -> proc_macro2::TokenS
...
                 meta
             }
         }
     } else {
         quote! {
             #i => {
                 let mut meta = miniconf::Metadata::default();
-
-                meta.max_length = stringify!(#field_name).len();
+                meta.max_length = 1 + stringify!(#field_name).len();
```
ryan-summers, line 131: Should outline where the `+1` is coming from. I think it's the prefix separator? But also, is separator guaranteed to be a single `char`?

miniconf_derive/src/lib.rs
```diff
@@ -182,60 +170,40 @@ fn derive_struct(
...
-                    #[allow(unreachable_code)]
-                    {
-                        // If a deferred field is done, strip off the field name again,
-                        // and advance to the next field.
-                        path.truncate(original_length);
-
-                        state[0] += 1;
-                        state[1..].fill(0);
-                    }
+            ) -> Result<usize, miniconf::IterError> {
```
ryan-summers, line 198: Not really following what this return value is supposed to be. Is it in the docs of the trait now?

miniconf_derive/src/lib.rs
```diff
@@ -182,60 +170,40 @@ fn derive_struct(
...
-
-                    // Note(unreachable) Without any deferred fields, every arm above returns
-                    #[allow(unreachable_code)]
-                    {
-                        // If a deferred field is done, strip off the field name again,
-                        // and advance to the next field.
-                        path.truncate(original_length);
-
-                        state[0] += 1;
-                        state[1..].fill(0);
```
ryan-summers, line 231: I'm not understanding what this refactor is doing. Don't we still need to zero out the `state` in some cases?

src/array.rs
```diff
@@ -136,43 +139,30 @@ impl<T: Miniconf, const N: usize> Miniconf for Array<T, N> {
     fn metadata() -> Metadata {
         let mut meta = T::metadata();
 
-        // Unconditionally account for separator since we add it
-        // even if elements that are deferred to (`Options`)
-        // may have no further hierarchy to add and remove the separator again.
-        meta.max_length += digits(N) + 1;
```
ryan-summers, line 142: This comment still seems relevant?

src/lib.rs
```diff
@@ -246,13 +251,13 @@ where
 
     fn set(&mut self, path: &str, data: &[u8]) -> Result<usize, Error> {
         let mut de = serde_json_core::de::Deserializer::new(data);
-        self.set_path(&mut path.split(Self::SEPARATOR).peekable(), &mut de)?;
+        self.set_path(&mut path.split(Self::SEPARATOR).skip(1), &mut de)?;
```
ryan-summers, line 254: Would be useful to document what the skip is for (preceding prefix). Alternatively, we _could_ make it accept both a preceding prefix and only pop it off if it's present? Seems like a good idea to me.

ryan-summers (changes requested):
Looking very intriguing to remove external deps. I really like the usability upgrades :)

miniconf_derive/src/lib.rs
jordens: It's wrong. We simply always add it. And that's now a redundant statement as any operation on the path anywhere starts with adding the separator. And the statement about removing the separator is wrong as we never need to remove the separator anymore.

miniconf_derive/src/lib.rs
ryan-summers: I see now, the trait got updated.

miniconf_derive/src/lib.rs
jordens: Yes. It's a single `char`.
But the length of a single char is not neccessarily 1.

jordens: deferred to #152

src/lib.rs
jordens: I prefer the simple control flow and unambiguous requirements.
Not skipping and instead stripping breaks e.g. the case of a bare `Option` (see `tests::option::just_option`).

miniconf_derive/src/lib.rs
jordens: This is done in `MiniconfIter::next()`. `next_path()` doesn't need to modify the state anymore at all.

miniconf_derive/src/lib.rs
ryan-summers: This should be as easy as calling `Self::Separator.len_utf8()` - don't know if we should really defer that?

src/array.rs
jordens: The explanation is now much simpler. Added comment.

miniconf_derive/src/lib.rs
ryan-summers: Ah, it's because the derive doesn't know anything about the separator, so we need to add the number of "Separators" into the metadata later, which we then use later to calculate if we have enough buffer etc. for a given serde impl

CHANGELOG.md
```diff
@@ -19,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   to provide the previously existing functionality.
 * The only required change for most downstream crates to adapt to the above is to
   make sure the `SerDe` trait is in scope (`use miniconf::SerDe`).
+* Paths now start with the path separator (unless they are empty).
```
ryan-summers, line 23: I think we need to note a change in miniconf_dervie as well. We'll need to republish both crates when we eventually release.

ryan-summers (approved):
LGTM, just need to note a new changelog entry for the derive crate

CHANGELOG.md
jordens: `miniconf_derive` doesn't have a changelog or a readme. I think it's usually done like that. derive crates are included, described and tracked in the main user-facing crate.

ryan-summers: I just want something to remind me that we can't relese until the derive crate is out, but I _think_ that will already happen because we changed the API of the trait

PR #156, opened 2023-07-25
be generic about the thing to write the path into

jordens:
The downside is that we don't check capacity anymore.
For most things that are `Write` they don't have a capacity.
On the other hand this was always done redundantly also in the mqtt client.

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -1,13 +1,11 @@
-use super::{IterError, Metadata, Miniconf, SerDe};
+use super::{IterError, Metadata, SerDe};
 use core::{fmt::Write, marker::PhantomData};
 
 /// An iterator over the paths in a Miniconf namespace.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub struct MiniconfIter<M: ?Sized, const L: usize, S, P> {
+pub struct MiniconfIter<M, S, const L: usize, P> {
     /// Zero-size marker field to allow being generic over M and gaining access to M.
-    miniconf: PhantomData<M>,
-    spec: PhantomData<S>,
-    path: PhantomData<P>,
+    marker: PhantomData<(M, S, P)>,
 
     /// The iteration state.
     ///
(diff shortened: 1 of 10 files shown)


tests/iter.rs
```diff
@@ -21,10 +21,10 @@ fn insufficient_space() {
     assert_eq!(meta.count, 3);
 
     // Ensure that we can't iterate if we make a state vector that is too small.
-    assert!(Settings::iter_paths::<1, 256>().is_err());
+    assert!(Settings::iter_paths::<1, String<256>>().is_err());
 
     // Ensure that we can't iterate if the topic buffer is too small.
-    assert!(Settings::iter_paths::<10, 1>().is_err());
+    // assert!(Settings::iter_paths::<10, String<1>>().is_err());
```
ryan-summers, line 27:
```suggestion
    assert!(Settings::iter_paths::<10, String<1>>().is_err());
```

ryan-summers: Or is the problem that we no longer detect this as an error? That seems like a problem

src/lib.rs
```diff
@@ -198,12 +198,11 @@ pub trait SerDe<S>: Miniconf {
     ///
     /// # Generics
     /// * `L`  - The maximum depth of the path, i.e. number of separators plus 1.
-    /// * `TS` - The maximum length of the path in bytes.
+    /// * `P`  - The type to hold the path.
     ///
     /// # Returns
     /// A [MiniconfIter] of paths or an [IterError] if `L` or `TS` are insufficient.
```
ryan-summers, line 204:
```suggestion
    /// A [MiniconfIter] of paths or an [IterError] if `P` or `TS` are insufficient.
```

src/iter.rs
```diff
@@ -24,26 +22,19 @@ pub struct MiniconfIter<M: ?Sized, const L: usize, const TS: usize, S> {
...
     }
 }
 
-impl<M: ?Sized + Miniconf + SerDe<S>, const L: usize, const TS: usize, S>
-    MiniconfIter<M, L, TS, S>
-{
+impl<M: SerDe<S>, S, const L: usize, P> MiniconfIter<M, S, L, P> {
     pub fn metadata() -> Result<Metadata, IterError> {
         let meta = M::metadata(M::SEPARATOR.len_utf8());
-        if TS < meta.max_length {
```
ryan-summers, line 43: If we can't pre-emptively check the length, do we even want to precalculate the data length? I guess it's still potentially useful to the end user for initial setup.

tests/iter.rs
jordens: See the PR description.
I think it's ok to skip it as we've always found it necessary to do a similar check and redundant check elsewhere as well.

ryan-summers: Oh sorry, I thought this was related to the first iteration (next()) not returning an error. The construction not returning an error is fine. Can we remove the broken test?

ryan-summers (approved):

PR #150, opened 2023-07-25
rj/inner

jordens:
- be generic over the serde format
- no unwrap
- factor out MiniconfJson trait
- pass separator as argument
- use marker struct and trait for separator
- move separator to spec trait
- rename MiniconfSpec -> SerDe
- remove marker trait
- use git serde-json-core
- add changelog entries
- shorten paths
- move metadata checking to iterator
- update docs
- style
- simplify ser lifetimes
- return serializer Ok
- simplify deserializer lifetimes
- simpler default
- cleanup
- refactor path building/parsing
- fixups
- remove peekable
- add test
- more changelog
- refactor
- serde-json-core git
- style
- remove old comments
- fix comment
- add Miniconf Inner/Outer type parameter and use
- remove Array, wip

PR #158, opened 2023-07-25
deps

jordens:
- reorg features
- syn 2


Cargo.toml
```diff
@@ -12,20 +10,22 @@ repository = "https://github.com/quartiq/miniconf"
...
 
 [dependencies]
 miniconf_derive = { path = "miniconf_derive" , version = "0.6" }
-serde-json-core = { git = "https://github.com/rust-embedded-community/serde-json-core.git" }
+serde-json-core = { git = "https://github.com/rust-embedded-community/serde-json-core.git" , optional = true}
 serde = { version = "1.0.120", features = ["derive"], default-features = false }
-log = "0.4"
-heapless = { version = "0.7", features = ["serde"] }
+log = {version = "0.4", optional = true}
+heapless = { version = "0.7", features = ["serde"], optional=true }
```
ryan-summers, line 20: We can remove the public re-exports on `heapless` and `serde_json_core` now I think

jordens: The exports are already feature-dependent.

ryan-summers: Yeah, what I'm saying is that they should be removed entirely now. They're unnecessary and should be deprecated/removed from what I understand.

jordens: Right. I'm not entirely sure about the serde_json_core errors though as they are still part of the API. OTOH they are accessible.

miniconf_derive/src/field.rs
```diff
@@ -8,14 +7,21 @@ pub struct StructField {
...
-            .filter(|attr| attr.path.is_ident("miniconf"))
-            .map(|attr| AttributeParser::new(attr.tokens.clone()).parse())
-            .collect();
+        let mut deferred = false;
 
-        let deferred = attributes.iter().any(|x| *x == MiniconfAttribute::Defer);
+        for attr in field.attrs.iter() {
+            if attr.path().is_ident("miniconf") {
+                attr.parse_nested_meta(|meta| {
+                    if meta.path.is_ident("defer") {
```
ryan-summers, line 15: The existing style, while some code overhead, was intended to make it easy to add new attributes while making it relatively maintainable. This is less code, but makes it harder to track new attributes in the future

ryan-summers (approved):

miniconf_derive/src/field.rs
jordens: It's just another `if`, i.e. about five lines.
Same amount as in the old code (c.f. https://github.com/quartiq/miniconf/pull/150/files#diff-1857dc627a1f48fd67d28c8ed2d35e53994fe4b9d71ee181c034cc98a30bd163)

PR #159, opened 2023-07-26
std json

jordens:
- remove heapless dependency from json feature
- rename features, add std json
- [x] depends on https://github.com/quartiq/minimq/issues/132 for `--all-features`


jordens:
The downside to the `Serde<Style>` thing is that once you have multiple enabled at the same time (e.g. `SerDe<SCPI>` and `SerDe<JsonCoreSlash>` for different simultaneous interfaces), you have to explicitly disambiguate each and every call to the `SerDe` trait.

ryan-summers (approved):

PR #160, opened 2023-07-27
add `traverse_by_key` and `{get,set}_by_key` methods and refactor several things

### LINKED ISSUE #155: shrink dependencies
* we have both `machine` and `smlang`
* move the `Serde<JsonCoreSlash>` behind a feature flag together with the `serde-json-core` dependency
* `log`, `smlang` and `minimq` are only used by `mqtt`?

jordens:
* remove the `SerDe<Style>` trait in favor of explicit and flexible individual traits
* add closure-based traversal of the namespaces either by index or by name.
* make ser/de generic over the type of key iterator. supported currently are indices and names.
* make the traversal generally more powerful indicating where Errors or Ok originated
* remove the separator from metadata, add helper method to account for it.

close #155


miniconf_derive/src/lib.rs
```diff
@@ -222,6 +230,27 @@ fn derive_struct(
...
+                P::Item: miniconf::Key,
+                F: FnMut(miniconf::Ok, usize, &str) -> Result<(), E>,
+            {
+                match keys.next() {
+                    None => Ok(miniconf::Ok::Internal(0)),
+                    Some(key) => {
+                        let index = miniconf::Key::find::<Self>(key).ok_or(miniconf::Error::NotFound(1))?;
+                        match index {
+                            #(#traverse_by_key_arms ,)*
+                            _ => Err(miniconf::Error::NotFound(1)),
```
ryan-summers, line 249: I'm not really following why the indices are being used. I also am concerned about consistency. Can we maybe add an impl to `Error` and make these use a default-style constructor, since the various types are always newly constructed with the same initial value? Something like:

```rs
miniconf::Error::not_found()
```

The idea being that the index provided is constant. I'm not sure if this is necessarily true, but in general it's really hard for me to understand what these indices are supposed to be initialized at, and would help if we didn't need to think about the complexity everywhere.

src/lib.rs
```diff
@@ -113,82 +131,158 @@ pub struct Metadata {
...
+            if ok == Ok::Leaf(0) {
+                Ok(())
+            } else {
+                path.write_str(sep).and_then(|_| path.write_str(name))
+            }
+        })
+    }
 
-    /// Get metadata about the paths in the namespace.
+    /// Convert `path` to `indices`.
```
ryan-summers: I do not understand what `path` is supposed to be? Are these docs outdated?

src/json_core.rs
```diff
@@ -1,29 +1,98 @@
...
+    }
+
+    fn set_json_by_index(
+        &mut self,
+        indices: &[usize],
+        data: &[u8],
+    ) -> Result<usize, Error<de::Error>> {
         let mut de = de::Deserializer::new(data);
-        self.set_path(&mut path.split(Self::SEPARATOR).skip(1), &mut de)?;
+        self.set_by_key(indices.iter().copied(), &mut de)?;
```
ryan-summers, line 85: Why do we have to copy the interator? Seems like we should restructure the function args instead because we don't use `indices` anywhere else

src/iter.rs
```diff
@@ -20,62 +21,80 @@ pub struct MiniconfIter<M, S, const L: usize, P> {
...
+where
+    M: Miniconf + ?Sized,
+    P: Write + Default,
+{
     type Item = P;
 
     fn next(&mut self) -> Option<Self::Item> {
         let mut path = Self::Item::default();
 
         loop {
```
ryan-summers, line 61: Why the loop if we're returning the very first M::path() value?

src/iter.rs
```diff
@@ -20,62 +21,80 @@ pub struct MiniconfIter<M, S, const L: usize, P> {
...
+                // Non-root leaf: advance at current depth
+                Ok(Ok::Leaf(depth)) => {
+                    self.count = self.count.map(|c| c - 1);
+                    self.state[depth - 1] += 1;
+                    Some(path)
                 }
-            }
+                // If we end at a leaf node, the state array is too small.
+                Ok(Ok::Internal(_depth)) => panic!("State too small"),
+                Err(e) => panic!("{e:?}"),
```
ryan-summers, line 96:
```suggestion
                Err(e) => panic!("Miniconf path iteration encountered an internal error: {e:?}"),
```

tests/arrays.rs
```diff
@@ -1,212 +1,137 @@
...
-    let mut s = S::default();
-
-    for _i in s.a.into_iter().flatten() {}
-
-    for _i in s.a.iter().flatten() {}
-
-    for _i in s.a.iter_mut().flatten() {}
+fn too_short() {
+    let mut s = Settings::default();
+    assert_eq!(s.set_json("/d", b"[1,2,3]"), Err(Error::Internal(1)));
```
ryan-summers, line 50:
```suggestion
    assert_eq!(s.set_json("/d", b"[1,2]"), Err(Error::Internal(1)));
```

tests/index.rs
```diff
@@ -0,0 +1,116 @@
...
+    let mut s = Settings::default();
+    s.set_json_by_index(&[0], b"[1,2]").unwrap();
+    assert_eq!(s.a, [1, 2]);
+}
+
+#[test]
+fn defer() {
+    let mut s = Settings::default();
+    s.set_json_by_index(&[1, 1], b"99").unwrap();
+    assert_eq!(s.d[1], 99);
```
ryan-summers, line 35: Would be worth testing that none of the other params of `s` changed.

ryan-summers (commented):
These are some very intriguing changes that make miniconf _way_ more flexible. I'm very interested to see. Is the intent with the `key` refactor to allow arbitrary path/separator syntaxes to be used without affecting the fundamental operation of miniconf?

src/lib.rs
ryan-summers: This got fixed in a new commit

src/lib.rs
```diff
@@ -32,70 +31,88 @@ pub use serde; // re-export
...
+    /// The names/indices are valid, but there is trailing data at the end.
+    TooLong(usize),
 
-    /// The value provided could not be serialized or deserialized.
-    ///
-    /// Check that the serialized data is valid and of the correct type.
-    /// Check that the buffer had sufficient space.
-    SerDe(T),
+    /// The names/indices are valid, but end early and do not reach a leaf.
+    Internal(usize),
```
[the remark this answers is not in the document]
jordens: Wikipedia calls them internal.
https://en.wikipedia.org/wiki/Tree_(data_structure)#:~:text=An%20internal%20node%20(also%20known,does%20not%20have%20child%20nodes.

miniconf_derive/src/lib.rs
```diff
@@ -42,95 +42,95 @@ pub fn derive(input: TokenStream) -> TokenStream {
...
-    let field_name = &struct_field.field.ident;
-    if struct_field.deferred {
+    if struct_field.defer {
         quote! {
             #i => {
-                let mut meta = <#field_type>::metadata(separator_length);
-                meta.max_length += separator_length + stringify!(#field_name).len();
-                meta.max_depth += 1;
-                meta
+                func(miniconf::Ok::Internal(1), #i, Self::NAMES[#i])?;
```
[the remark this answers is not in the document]
jordens: The Miniconf trait.
`traverse_by_key(keys, func: F) where F ...`

src/lib.rs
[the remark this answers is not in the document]
ryan-summers: Yeah, I guess when I see `Error::Internal`, it doesn't make me think of nodes in the settings tree, it makes me think that miniconf hit internal issues.

Maybe `Error::InternalNode`?

miniconf_derive/src/lib.rs
```diff
@@ -156,53 +156,61 @@ fn derive_struct(
...
-                D: miniconf::serde::Deserializer<'b>,
+                P: Iterator,
+                D: miniconf::serde::Deserializer<'a>,
+                P::Item: miniconf::Key,
             {
-                match path_parts.next().ok_or(miniconf::Error::PathTooShort)? {
-                    #(#set_path_arms ,)*
-                    _ => Err(miniconf::Error::PathNotFound),
+                let key = keys.next().ok_or(miniconf::Error::Internal(0))?;
+                let index = miniconf::Key::find::<Self>(key).ok_or(miniconf::Error::NotFound(1))?;
```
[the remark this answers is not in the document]
ryan-summers: I'm primarily not understanding the purpose of the `Key` trait I guess

miniconf_derive/src/lib.rs
```diff
@@ -156,53 +156,61 @@ fn derive_struct(
...
-                P: Iterator<Item = &'a str>,
-                D: miniconf::serde::Deserializer<'b>,
+                P: Iterator,
+                D: miniconf::serde::Deserializer<'a>,
+                P::Item: miniconf::Key,
             {
-                match path_parts.next().ok_or(miniconf::Error::PathTooShort)? {
-                    #(#set_path_arms ,)*
-                    _ => Err(miniconf::Error::PathNotFound),
+                let key = keys.next().ok_or(miniconf::Error::Internal(0))?;
```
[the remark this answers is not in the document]
jordens: Not here. If there is no further key, we are at the root with depth zero.
The rule is always: "each time we get something with next, the depth increases by one".

miniconf_derive/src/lib.rs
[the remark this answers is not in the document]
jordens: Everything that is `Key` can be used. I.e., and most importantly also `&str`.
All without any additional code.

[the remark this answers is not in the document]
ryan-summers: Ah okay I get you, so you can theoretically index into the settings structure using a key-like object, which could either be a path, a slice of strings, or a slice of indices. That makes sense.

miniconf_derive/src/lib.rs
jordens: I'd like to be explicit here. We are at depth = 1. This needs to be there because we got `Some()` from `next()`. It's distinct from the `Internal(0)` above. If we had `not_found()` and `internal()` that would obscure it for me.

src/json_core.rs
jordens: This is not copying the iterator but turning `Iterator<&usize>` into `Iterator<usize>`.

src/iter.rs
jordens: We are not. There is exactly one `continue` branch.

ryan-summers: Ah yes, I see now, thanks!

src/iter.rs
jordens: Hmm. That doesn't help me. The `Miniconf path iteration` context is better read from the backtrace and the panic message and all `panics` should be internal errors. A panic due to an non-internal error would be an obvious design error.
I did think about this and really didn't see what meaningful information I could add in the message.

tests/arrays.rs
jordens: This also tests precedence of `Internal` over `Inner`

src/lib.rs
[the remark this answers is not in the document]
jordens: Let's call it `TooShort` here as that's more descriptive of the error, and keep `Ok::Internal` below.

src/iter.rs
ryan-summers: My main point is for things where the panic trace information is lost (i.e. production). A good example is `booster`'s `service` command. Literally all we have is the serialized `Panic` contents - if we just do a debug print of the error, we don't even know that it relates to `Miniconf`

tests/arrays.rs
ryan-summers: Ah, gotcha, I didn't realize that when reviewing. Can you add a note? It looks like a bug otherwise.

jordens: Made it explicit.

src/iter.rs
jordens: Ack. I added a bit of context to both messages.

ryan-summers (approved):

PR #161, opened 2023-07-29
Miscellaneous 

jordens:
* `path` and `indices` should be generic over `keys: IntoIterator`
* `indices` can take any `Iterator<Item=&'a mut usize>`
* clarify code
* add more tests
* remove the leaf argument from the callback
* remove the Result type alias
* remove the Ok type in favor of plain usize
* make the iterator return Result
* remove the last panic from the core lib
* re-enable accidentally disabled tests
* add support for tuple structs
* reorder methods for consistency


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * Traversal by names or indices has been added through `Miniconf::traverse_by_key()`.
+* The `Miniconf` derive macro supports (unnamed) tuple structs.
```
ryan-summers, line 12: Can you update the `Limitations` section of the README as well? It says tuple structs aren't supported

src/iter.rs
```diff
@@ -96,6 +96,10 @@ where
                 }
                 // If we end at a leaf node, the state array is too small.
                 Err(e @ (Error::TooShort(_) | Error::Inner(_))) => Some(Err(e)),
+                // Not having consumed any name/index, the only possible case here is a bare option.
+                // And that can not return `NotFound`.
+                // Err(Error::NotFound(0)) => unreachable!(),
```
ryan-summers, line 101:
```suggestion
```

tests/tuple.rs
```diff
@@ -0,0 +1,18 @@
...
+    struct Settings(i32, f32);
+
+    let mut s = Settings::default();
+
+    let mut buf = [0u8; 256];
+    let len = s.get_json("/0", &mut buf).unwrap();
+    assert_eq!(&buf[..len], b"0");
+
+    s.set_json("/1", b"3.0").unwrap();
+    assert_eq!(s.1, 3.0);
```
ryan-summers, line 17: Mind adding a test that something like `/foo` and `/2` both fail using `set_json()`?

ryan-summers (changes requested):
Looking nice, cool to see how trivial it was to add support for tuple structs.

ryan-summers:
Ah, was just looking at a few commits. Let me re-review.

src/lib.rs
```diff
@@ -40,26 +40,26 @@ pub use serde::{de::DeserializeOwned, Serialize};
...
     Absent(usize),
 
-    /// The names/indices are valid, but there is trailing data at the end.
-    TooLong(usize),
-
-    /// The names/indices are valid, but end early and do not reach a leaf.
+    /// The key ends early and does not reach a leaf node.
     TooShort(usize),
 
+    /// A key was not found, is too large or invalid.
```
ryan-summers, line 57: I'm not sure what "Too large" is supposed to mean? Also, what's the difference between "invalid" and "not found"?

src/lib.rs
```diff
@@ -142,112 +127,120 @@ impl Metadata {
...
     ///
     /// Traversal is aborted once `func` returns an `Err(E)`.
     ///
-    /// May not exhaust the iterator if a leaf is found early. i.e. keys may be too long.
-    /// If `Self` is a leaf, nothing will be consumed from the iterator
-    /// and [`Ok::Leaf(0)`] will be returned.
-    /// If `Self` is non-leaf (internal) and the iterator is exhausted (empty),
-    /// [`Ok::Internal(0)`] will be returned.
+    /// May not exhaust `keys` if a leaf is found early. i.e. `keys`
+    /// may be longer than required.
```
ryan-summers, line 196: What's the design decision driving this? Seems odd to just ignore extra keys in the iterator when that would imply that we're traversing to something potentially not specified by the user.

Take for example one firmware version exposes the path: "/settings/enable"

Then, in a later version, firmware exposes "/settings/enable/power_disable".

If the user wrong to `/settings/enable/power_disable` on FW V1, would that mean that it resolves to "/settings/enable"?

src/lib.rs
```diff
@@ -260,68 +253,61 @@ pub trait Miniconf {
...
-    /// # Panic
-    /// `L` must be sufficiently large to hold the iterator state.
-    /// While this function will not panic itself, calling `Iterator::next()` on its
-    /// return value may.
-    ///
     /// # Generics
-    /// * `L`  - The maximum depth of the path, i.e. number of separators plus 1.
-    /// * `P`  - The type to hold the path.
+    /// * `L`  - The maximum depth of the path, i.e. the number of separators.
+    /// * `P`  - The type to hold the path. Needs to be `core::fmt::Write`.
```
ryan-summers, line 305: Ditto, trait requirement on P instead of docs?

src/array.rs
```diff
@@ -123,16 +136,19 @@ impl<T: Miniconf, const N: usize> Miniconf for Array<T, N> {
...
     {
         let key = keys.next().ok_or(Error::TooShort(0))?;
         let index = key.find::<Self>().ok_or(Error::NotFound(1))?;
-        let item = self.0.get(index).ok_or(Error::NotFound(1))?;
-        item.get_by_key(keys, ser).increment()
+        if index >= N {
+            return Err(Error::NotFound(1));
+        }
+        func(index, itoa::Buffer::new().format(index))?;
+        T::traverse_by_key(keys, func).increment()
```
ryan-summers, line 151: This kind of recursion might cause stack usage to get _really high_ if you have a large array of elements.

src/lib.rs
ryan-summers: It seems this isn't allowed for `get/set_by_key`, so I'm confused why it's the case here.

src/iter.rs
```diff
@@ -53,47 +56,51 @@ where
...
+                Ok(depth) => {
                     self.count = self.count.map(|c| c - 1);
                     self.state[depth - 1] += 1;
-                    Some(path)
+                    Some(Ok(path))
                 }
                 // If we end at a leaf node, the state array is too small.
-                Ok(Ok::Internal(_depth)) => panic!("Path iteration state too small"),
-                Err(e) => panic!("Path iteration: {e:?}"),
+                Err(e @ (Error::TooShort(_) | Error::Inner(_))) => Some(Err(e)),
```
ryan-summers, line 98: Isn't this guaranteed to not happen because we check `L` against the max depth specified by `metadata`? Seems like we could internally panic instead of forcing out a `Result` that the user has to unwrap that should never be `Err()` if miniconf is working properly.

src/option.rs
```diff
@@ -83,102 +83,102 @@ impl<T> From<Option<T>> for core::option::Option<T> {
...
         } else {
             Err(Error::Absent(0))
         }
     }
 
+    fn traverse_by_key<K, F, E>(_keys: K, _func: F) -> Result<usize, Error<E>>
+    where
+        F: FnMut(usize, &str) -> Result<(), E>,
+    {
+        Ok(0)
```
ryan-summers, line 175: Why aren't we calling the function for the single element contained in the option?

tests/basic.rs
```diff
@@ -22,11 +22,58 @@ fn meta() {
...
+    let f = |_, _: &_| unreachable!();
+    assert_eq!(
+        S::traverse_by_key([0].into_iter(), f),
+        Err(Error::<()>::NotFound(1))
+    );
+    assert_eq!(
+        S::traverse_by_key([0; 0].into_iter(), f),
+        Err(Error::TooShort(0))
+    );
+    assert_eq!(Option::<i32>::traverse_by_key([0].into_iter(), f), Ok(0));
```
ryan-summers, line 70: Should `traverse_by_key()` and friends take an `IntoIter`?

src/lib.rs
```diff
@@ -260,68 +253,61 @@ pub trait Miniconf {
...
     /// This is a depth-first walk.
-    /// The iterator will walk all paths, even those that may be absent at run-time (see [Option]).
+    /// The iterator will walk all paths, including those that may be absent at
+    /// run-time (see [Option]).
     /// The iterator has an exact and trusted [Iterator::size_hint].
     ///
     /// # Generics
     /// * `L`  - The maximum depth of the path, i.e. the number of separators.
-    /// * `P`  - The type to hold the path.
+    /// * `P`  - The type to hold the path. Needs to be `core::fmt::Write`.
```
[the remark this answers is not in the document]
ryan-summers: i.e `<P: core::fmt::Write>`

src/array.rs
jordens: It only recurses depth, not breadth.
The stack usage and recursion pattern is the same as before.

ryan-summers: Ah you're right. Then I'm not too concerned

src/iter.rs
jordens: It's a choice. If you do `iter_paths()` (`PathIter::new()`) then it checks and errors early.
If you do `iter_paths_unchecked()` (`PathIter::new_unchecked()`), you'll get `TooShort()` when iterating (before it just paniced).
In the mqtt client because of other constraints we check metadata explicitly, then use `_unchecked()` and unwrap in the iteration.

src/lib.rs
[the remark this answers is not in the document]
jordens: I think the error message will be the nearly the same in both cases.
Adding the bound would add more code, and more noise on refactoring.

src/iter.rs
ryan-summers: The intent of `PathIter::new_unchecked()` was that the operator of `miniconf` **guaranteed** that the state vector was long enough, which is why miniconf panicked internally for ergonomics. I'm not personally a fan of this API change when I don't see any utility in it

src/lib.rs
jordens: This is in line with the previous behaviour where the iterator state was allowed to be too long.
The alternative would not change much to the method: you'd get `Err(TooLong(depth))` instead of `Ok(depth)` back but the call pattern to `func()` would always be the same.
And it would make the match in `PathIter::next()` a bit more complicated (as we don't want to find the right depth for each `path()` call first).
The important thing is probably the sideeffect:
Not checking for too long paths here also allows the remaining path elements to be used for something else since the iterator can be passed by mutable reference and isn't consumed, (maybe like URLs?).

For `{get,set}_by_key` it's decidedly different. There I want the keys to be unambiguous and they can only ever meaningfully refer to a leaf and want to exhaust the iterator.

src/mqtt_client/mod.rs
```diff
@@ -59,7 +59,7 @@ mod sm {
             Self {
                 clock,
                 timeout: None,
-                republish_state: M::iter_paths("/").unwrap(),
+                republish_state: M::iter_paths_unchecked("/"),
```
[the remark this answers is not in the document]
jordens: This was never a compile time check, only ever at run-time early. (L217-219).
Doing the check here again is just redundant.

src/option.rs
jordens: It was wrong to call it before. We never call it here if we don't add any path hierarchy levels here.

tests/basic.rs
jordens: I thought long and hard and went back and forth on this. I don't think so:
For the recursive pattern (`*_by_key()`) we take Iterators, since IMO the code noise of doing `.into_iter()` at each level in the implementation outweighs the flexibility gained and outweighs the code noise of making the iterator at the callsite. Consequently the non-recursive methods that are much more frequently called by the user (`{get,set}_json()`, `path()`, `indices()`) do take `IntoIterator`.
In short it does seem like `IntoIterator` doesn't look so good for recursive consumption because of the pointless `into_iter()` at all lower recursion levels, while it is perfect for iterative consumption where you only do `into_iter()` once().

src/mqtt_client/mod.rs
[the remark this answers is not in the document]
ryan-summers: L217-L219 is about the topic length, not the iterator state depth. I agree we only need a `checked()` and `unwrap()` once, but now, it's not doing it anywhere

src/iter.rs
jordens: I don't see it that wal.

In the worst case this is moving panics (string too short, state too short) in a library to an unwrap (thus panic) in the application (or a higher layer as mqtt-client). There is no penalty for that.
In the best case these errors can now be handled in the application.

`_unchecked()` avoids gathering the complete metadata for each iteration an does not do the early check for state length.

src/lib.rs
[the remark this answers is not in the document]
jordens: Interestingly, adding the bound doesn't change or move the second error, it just adds another one before that.

```
error[E0277]: the trait bound `i32: std::fmt::Write` is not satisfied
   --> tests/iter.rs:76:44
    |
76  |     let i = Option::<i32>::iter_paths::<0, i32>("/").unwrap();
    |                                            ^^^ the trait `std::fmt::Write` is not implemented for `i32`
    |
    = help: the following other types implement trait `std::fmt::Write`:
              &mut W
              Formatter<'_>
              OsString
              core::net::display_buffer::DisplayBuffer<SIZE>
              miniconf::minimq::embedded_nal::heapless::String<N>
              miniconf::minimq::embedded_nal::heapless::Vec<u8, N>
              std::string::String
note: required by a bound in `iter_paths`
   --> /home/rj/src/miniconf/src/lib.rs:293:38
    |
293 |     fn iter_paths<const L: usize, P: core::fmt::Write>(
    |                                      ^^^^^^^^^^^^^^^^ required by this bound in `Miniconf::iter_paths`

error[E0599]: the method `next` exists for struct `PathIter<'_, Option<i32>, 0, i32>`, but its trait bounds were not satisfied
  --> tests/iter.rs:77:7
   |
77 |     i.next();
   |       ^^^^ method cannot be called on `PathIter<'_, Option<i32>, 0, i32>` due to unsatisfied trait bounds
   |
  ::: /home/rj/src/miniconf/src/iter.rs:6:1
   |
6  | pub struct PathIter<'a, M: ?Sized, const L: usize, P> {
   | ----------------------------------------------------- doesn't satisfy `_: Iterator`
   |
   = note: the following trait bounds were not satisfied:
           `i32: std::fmt::Write`
           which is required by `PathIter<'_, std::option::Option<i32>, 0, i32>: Iterator`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `miniconf` (test "iter") due to previous error
```

src/mqtt_client/mod.rs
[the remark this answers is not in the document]
jordens: ?
`        assert!(meta.max_depth <= MAX_RECURSION_DEPTH);`

src/lib.rs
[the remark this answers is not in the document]
ryan-summers: Maybe we should add the `Write` bound to the `PathIter` struct directly instead of the impl?

src/iter.rs
ryan-summers: The original design intent was that miniconf iteration would never be error-able. This change is reversing that design choice, and I don't really see why

src/lib.rs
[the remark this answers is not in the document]
jordens: I now added the bound to the `Miniconf` trait as well.
Duplicating it in the `struct` definition or the inherent impl doesn't seem necessary or helpful.

src/mqtt_client/mod.rs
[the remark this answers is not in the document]
ryan-summers: Ah, I was looking at `main` instead of this PR. I see now, thanks :)

src/iter.rs
ryan-summers: I guess my main point is that this introduces thrash to the end user and the miniconf code base, and I don't really see the purpose

jordens: Before the iteration could always panic but would never return an error.
There was support for checks ahead of time (still at runtime) to exclude these panics. Then the errors would arise before iterating. Now iteration never panics.
The checks are still there (the early string length check is in mqtt-client). The only change to the before-iteration checks is that the capacity check needs to be done by the user (since `Write` doesn't give capacity info from the type alone, unlike `heapless::String<N>`). On balance that change (moving the capacity check) should be clearly a positive one.
When choosing between an API that panics (before) and one that errors (now), I'd also prefer the later as that at least in principle allows handling the error.

ryan-summers (approved):

src/iter.rs
jordens: I agree that when converting to the new API, adding an `.unwrap()` to at the callsite adds some thrashing. But it is a gain in robustness. And AFAICT it's no meaningful difference in generated code in the current practical cases. For the user right now that only uses `MqttClient` there is zero change anyway.

PR #162, opened 2023-07-31
make miniconf trait generic over depth

jordens:
Removes the need for newtypes
Makes the iterator automatically know the required state size

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,14 +5,14 @@
 [![Continuous Integration](https://github.com/vertigo-designs/miniconf/workflows/Continuous%20Integration/badge.svg)](https://github.com/quartiq/miniconf/actions)
 
 Miniconf enables lightweight (`no_std`) partial serialization (retrieval) and deserialization
-(updates, modification) within a hierarchical namespace by path. The namespace is backed by
-structs and arrays of serializable types.
+(updates, modification) within a tree by key. The tree is backed by
+structs/arrays/Options of serializable types.
 
 Miniconf can be used as a very simple and flexible backend for run-time settings management in embedded devices
 over any transport. It was originally designed to work with JSON ([serde_json_core](https://docs.rs/serde-json-core))
 payloads over MQTT ([minimq](https://docs.rs/minimq)) and provides a comlete [MQTT settings management
 client](MqttClient) and a Python reference implementation to ineract with it.
-`Miniconf` is generic over the `serde::Serializer`/`serde::Deserializer` backend and the path hierarchy separator.
+`Miniconf` is completely generic over the `serde::Serializer`/`serde::Deserializer` backend and the path hierarchy separator.
 
 ## Example
 ```rust
@@ -129,24 +129,29 @@ python -m miniconf -d quartiq/application/+ foo=true
 
 ## Design
 For structs Miniconf offers a [derive macro](derive.Miniconf.html) to automatically
-assign a unique path to each item in the namespace of the struct.
+assign a unique path to each item in the tree.
 The macro implements the [Miniconf] trait that exposes access to serialized field values through their path.
-All types supported by [serde] (and the `serde::Serializer`/`serde::Deserializer` backend) can be used as fields.
+All types supported by [serde] (and the `serde::Serializer`/`serde::Deserializer` backend) or `Miniconf`
+can be used as fields.
 
-Elements of homogeneous [core::array]s are similarly accessed through their numeric indices.
-Structs, arrays, and Options can then be cascaded to construct a multi-level namespace.
-Namespace depth and access to individual elements instead of the atomic updates
-is configured at compile (derive) time using the `#[miniconf(defer)]` attribute.
-`Option` is used with `#[miniconf(defer)]` to support paths that may be absent (masked) at
-runtime.
+Elements of homogeneous [core::array]s can be made accessible either
+1. as a single leaf in the tree like other serde-capable items, or
+2. atomically through their numeric indices (with the attribute `#[miniconf(defer)]` or `#[miniconf(defer(1))]`), or
+3. through their sub-trees with `#[miniconf(defer(D))]` and `D >= 2`.
 
-The [Miniconf] implementations for [core::array] and [core::option::Option] provide
-atomic access to their respective inner element(s) or expose deep access
-into the inner element(s) through their respective inner [Miniconf] implementations.
+`Option` is used 
+1. like a standard `serde` Option, or
+2. with `#[miniconf(defer)]` to support paths that may be absent (masked) at runtime.
+3. with `#[miniconf(defer(D))]` and `D >= 2` to support masking entire sub-trees at runtime.
+
+Structs, arrays, and Options can then be cascaded to construct more complex trees.
+
+Lookup into the tree is done using an iterator over [Key] items. `usize` indices or `&str` names are supported.
+
+Path iteration is supported with arbitrary separator between names.
 
 ## Formats
-Miniconf is generic over the `serde` backend/payload format and the path hierarchy separator
-(as long as the path can be split by it unambiguously).
+Miniconf is generic over the `serde` backend/payload format and the path hierarchy separator.
 
 Currently support for `/` as the path hierarchy separator and JSON (`serde_json_core`) is implemented
 through the [JsonCoreSlash] style.
@@ -156,8 +161,8 @@ Miniconf is designed to be protocol-agnostic. Any means that can receive key-val
 some external source can be used to modify values by path.
 
 ## Limitations
-Deferred (non-atomic) access to inner elements of some types is not yet supported, e.g. enums
-other than [core::option::Option].
+Deferred/deep/non-atomic access to inner elements of some types is not yet supported, e.g. enums
+other than [core::option::Option]. These are still however usable in their atomic `serde` form as leaves.
 
 ## Features
 * `mqtt-client` Enable the MQTT client feature. See the example in [MqttClient].
--- a/README.md
+++ b/README.md
@@ -49,18 +49,18 @@ struct Settings {
     #[miniconf(defer)]
     array_defer: [i32; 2],
     // ... or deferring to two levels (index and then inner field name)
-    #[miniconf(defer)]
-    array_miniconf: miniconf::Array<Inner, 2>,
+    #[miniconf(defer(2))]
+    array_miniconf: [Inner; 2],
 
     // Hiding paths by setting the Option to `None` at runtime
     #[miniconf(defer)]
     option_defer: Option<i32>,
     // Hiding a path and deferring to the inner
-    #[miniconf(defer)]
-    option_miniconf: miniconf::Option<Inner>,
+    #[miniconf(defer(2))]
+    option_miniconf: Option<Inner>,
     // Hiding elements of an Array of Miniconf items
-    #[miniconf(defer)]
-    array_option_miniconf: miniconf::Array<miniconf::Option<Inner>, 2>,
+    #[miniconf(defer(3))]
+    array_option_miniconf: [Option<Inner>; 2],
 }
 
 let mut settings = Settings::default();
--- a/README.md
+++ b/README.md
@@ -53,13 +53,13 @@ struct Settings {
     array_miniconf: [Inner; 2],
 
     // Hiding paths by setting the Option to `None` at runtime
-    #[miniconf(defer)]
+    #[miniconf(defer(0))]
     option_defer: Option<i32>,
     // Hiding a path and deferring to the inner
-    #[miniconf(defer(2))]
+    #[miniconf(defer(1))]
     option_miniconf: Option<Inner>,
     // Hiding elements of an Array of Miniconf items
-    #[miniconf(defer(3))]
+    #[miniconf(defer(2))]
     array_option_miniconf: [Option<Inner>; 2],
 }
 
@@ -97,7 +97,7 @@ let len = settings.get_json("/struct_", &mut buf)?;
 assert_eq!(&buf[..len], br#"{"a":3,"b":3}"#);
 
 // Iterating over all paths
-for path in Settings::iter_paths::<3, String>("/").unwrap() {
+for path in Settings::iter_paths::<String>("/") {
     let path = path.unwrap();
     // Serializing each
     match settings.get_json(&path, &mut buf) {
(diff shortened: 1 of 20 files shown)


src/array.rs
```diff
@@ -1,98 +1,19 @@
...
+/// one atomic value.
 ///
-/// For standard arrays of [`[T; N]`](array) form, by default the entire array is accessed as one atomic
-/// value. By adding the `#[miniconf(defer)]` attribute, each index of the array is is instead accessed as
-/// one atomic value (i.e. a single Miniconf item).
-///
-/// The type you should use depends on what data is contained in your array. If your array contains
-/// `Miniconf` items, you can (and often want to) use [`Array`] and the `#[miniconf(defer)]` attribute.
+/// The type to use depends on what data is contained in your array. If your array contains
+/// `Miniconf` items, you can (and often want to) use `D >= 2`.
```
ryan-summers, line 14: It's not really clear to me what a `D = 3` would even do? I guess that's further deferrals through options etc? Maybe it's worth noting that trivial `Miniconf` items such as structs just ignore excessive deferral.

src/lib.rs
```diff
@@ -130,33 +130,35 @@ impl Metadata {
...
+/// The parameter `Y` is a lower bound for the recursion depth. It's only non-unity in
+/// `array` and `option` as they need a configurable recursion depth at compile time
+/// and can't be marked up with inner attributes.
+/// The paramter `Y` deviates from the key tree depth significantly:
+/// * `Option` consume a layer of Miniconf recursion, but don't add to key tree height
+/// * A struct that has `Miniconf` derived for it will only implement `Miniconf<1>`
+///   but may perform arbitrarily recursion and also have arbitrarily long keys.
+///   `1` is then the minimum recursion depth for a (non-newtype) struct. No other
+///   value is needed or valid.
+pub trait Miniconf<const Y: usize> {
```
ryan-summers, line 161: I think we could make this default to 1, since that's the "normal" operational mode.
```suggestion
pub trait Miniconf<const Y: usize = 1> {
```

src/mqtt_client/mod.rs
```diff
@@ -21,7 +21,7 @@ const MAX_RECURSION_DEPTH: usize = 8;
 // republished.
 const REPUBLISH_TIMEOUT_SECONDS: u32 = 2;
 
-type Iter<M> = crate::PathIter<'static, M, MAX_RECURSION_DEPTH, String<MAX_TOPIC_LENGTH>>;
+type Iter<M> = crate::PathIter<'static, M, MAX_RECURSION_DEPTH, String<MAX_TOPIC_LENGTH>, 1>;
```
ryan-summers, line 24: I'd say that ideally all of these excess `1` params could be wiped with generic defaults.

ryan-summers (commented):
This is absolutely fascinating to review, and solves the `Array` and `Option` problem extremely nicely. Nice work! This seems like a really nice clean up to usability.

I have some minor suggestions to remove a lot of unnecessary genericism.

miniconf_derive/src/field.rs
```diff
@@ -1,19 +1,26 @@
...
                 attr.parse_nested_meta(|meta| {
+                    if meta.input.is_empty() {
+                        return Ok(());
+                    }
                     if meta.path.is_ident("defer") {
-                        defer = true;
+                        let content;
+                        parenthesized!(content in meta.input);
+                        let lit: LitInt = content.parse().unwrap();
+                        defer = Some(lit.base10_parse().unwrap());
```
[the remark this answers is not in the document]
jordens: That'll just get unwrapped later in L29.
These internal panics are how we've always done it. And the error messages on these panics are meant for user consumption. This snippet is pretty much right from the `syn` example.
Proper diagnostics would involve showing the span of the error etc.

src/array.rs
jordens: e.g. these:

```
#[miniconf(defer(3))]
a: Option<[[i32; 2]; 2]>,
#[miniconf(defer(4))]
a: Option<[Option<[i32; 2]>; 2]>,
```

jordens: They don't ignore excessive deferral. That'll give you a trait error!

src/lib.rs
jordens: Ah. Right! Even better

src/mqtt_client/mod.rs
jordens: That one can't be remove since it's on a type, not a trait. But there were a bunch of others.

src/array.rs
jordens: A struct (that has Miniconf derived) can only be used with `#[miniconf(defer(1))]` or no miniconf at all. That's why it's lower bound.

jordens: The awkward interpretation of the trait parameter is resolved by #170 . It's now always just the maximum depth.

README.md
```diff
@@ -49,18 +49,18 @@ struct Settings {
     #[miniconf(defer)]
     array_defer: [i32; 2],
     // ... or deferring to two levels (index and then inner field name)
-    #[miniconf(defer)]
-    array_miniconf: miniconf::Array<Inner, 2>,
+    #[miniconf(defer(2))]
```
ryan-summers, line 52: WOuld the term `depth` make more sense than `defer` now? Both technically work, just a thought.

src/array.rs
```diff
@@ -1,98 +1,21 @@
...
-}
-
-impl<T, const N: usize> From<Array<T, N>> for [T; N] {
-    fn from(x: Array<T, N>) -> Self {
-        x.0
-    }
-}
+/// of `u32`), then you must use `D = 1` or `D = 0` if all items are to be accessed simultaneously.
+/// For e.g. `[[T; 2]; 3] where T: Miniconf<3>` you may want to use `D = 5` (note that `D <= 2`
+/// will also work if `T: Serialize + DeserializeOwned`).
```
ryan-summers, line 18: This feels unfortunate to me because it means the user of miniconf then needs to somehow know the internal depth of some struct where they called `#[derive(Miniconf)]` and the depth was auto-calculated for them.

If I were writing i.e. Booster's settings tree, I'm not sure what depth I should actually use for the various channel settings arrays without having to do a relatively complex analysis. Is there any way we could save the user from this and make it "Just work"?

src/iter.rs
```diff
@@ -26,34 +26,32 @@ pub struct PathIter<'a, M: ?Sized, const L: usize, P> {
...
-            return Err(SliceShort);
-        }
+        assert!(Y == meta.max_depth);
         let mut s = Self::new_unchecked(separator);
         s.count = Some(meta.count);
-        Ok(s)
+        s
     }
 
     pub(crate) fn new_unchecked(separator: &'a str) -> Self {
```
ryan-summers, line 41: I don't see any reason to keep the `new_unchecked()` API any more now that we guarantee, at compile time, the depth equivalency to the type parameter. It seems to me that we should have a single `new()`.

src/lib.rs
```diff
@@ -130,33 +130,27 @@ impl Metadata {
 /// Capability to convert a key into an node index.
 pub trait Key {
     /// Convert the key `self` to a `usize` index.
-    fn find<M: Miniconf>(&self) -> core::option::Option<usize>;
+    fn find<const D: usize, M: Miniconf<D>>(&self) -> core::option::Option<usize>;
```
ryan-summers, line 133: Can this also have a default value for the const-generic? Not sure if rust supports that in the position.

src/mqtt_client/mod.rs
```diff
@@ -139,7 +138,7 @@ impl<'a> Command<'a> {
 ///     foo: bool,
 /// }
 ///
-/// let mut client: MqttClient<Settings, _, _, 256> = MqttClient::new(
+/// let mut client: MqttClient<_, _, _, 256, 1> = MqttClient::new(
```
ryan-summers, line 141:
```suggestion
/// let mut client: MqttClient<_, _, _, 256, _> = MqttClient::new(
```

Or does this not work?

src/option.rs
```diff
@@ -16,120 +13,65 @@ use core::ops::{Deref, DerefMut};
...
-        K::Item: Key,
-        S: serde::Serializer,
-    {
-        if let Some(inner) = self.0.as_ref() {
-            inner.get_by_key(keys, ser)
-        } else {
-            Err(Error::Absent(0))
+/// If the depth specified by the `miniconf(defer(D))` attribute exceeds 0,
+/// the `Option` can be used to access content within the inner type.
+/// If marked with `#[miniconf(---)]`, and `None` at runtime, the value or the entire sub-tree
```
ryan-summers, line 18:
```suggestion
/// If marked with `#[miniconf(defer(-))]`, and `None` at runtime, the value or the entire sub-tree
```

tests/arrays.rs
```diff
@@ -109,31 +109,24 @@ fn iter() {
...
-    assert!(
-        miniconf::Array::<miniconf::Array<S, 0>, 0>::iter_paths::<2, String>("")
-            .unwrap()
-            .next()
-            .is_none()
-    );
 
     #[derive(Miniconf)]
     struct Q {
         #[miniconf(defer)]
```
ryan-summers, line 126:
```suggestion
        #[miniconf(defer(2))]
```

I don't think the change to the test is equivalent, not sure if 2 is the correct value.

tests/integration_test.rs
```diff
@@ -102,7 +102,7 @@ fn main() -> std::io::Result<()> {
     .unwrap();
 
     // Construct a settings configuration interface.
-    let mut interface = miniconf::MqttClient::<Settings, _, _, 256>::new(
+    let mut interface = miniconf::MqttClient::<_, _, _, 256, 2>::new(
```
ryan-summers, line 105:
```suggestion
    let mut interface = miniconf::MqttClient::<_, _, _, 256, _>::new(
```

Would prefer if the user had no knowledge of Miniconf Depth

tests/option.rs
```diff
@@ -102,14 +102,14 @@ fn option_test_normal_option() {
 fn option_test_defer_option() {
     #[derive(Copy, Clone, Default, Miniconf)]
     struct S {
-        #[miniconf(defer)]
+        #[miniconf(defer(0))]
```
ryan-summers, line 105: Isn't defer(0) the same as not putting an attribute at all? Or is this only different for `Option`?

tests/validation_failure.rs
```diff
@@ -82,7 +82,7 @@ async fn main() {
     tokio::task::spawn(async move { client_task().await });
 
     // Construct a settings configuration interface.
-    let mut interface: miniconf::MqttClient<Settings, _, _, 256> = miniconf::MqttClient::new(
+    let mut interface: miniconf::MqttClient<_, _, _, 256, 1> = miniconf::MqttClient::new(
```
ryan-summers, line 85:
```suggestion
    let mut interface: miniconf::MqttClient<_, _, _, 256, _> = miniconf::MqttClient::new(
```

ryan-summers (commented):
In general, I like this change. However, I'm concerned that the `defer()` attribute requires the user to have knowledge of derived-depth when affixing the `defer()` attribute.

In reality, the user probably doesn't know anything about depth, and I'd prefer if they didn't have to think about it if possible.

examples/mqtt.rs
```diff
@@ -78,7 +78,7 @@ async fn main() {
     // Spawn a task to send MQTT messages.
     tokio::task::spawn(async move { mqtt_client().await });
 
-    let mut client: MqttClient<Settings, Stack, StandardClock, 256> = MqttClient::new(
+    let mut client: MqttClient<_, _, _, 256, 2> = MqttClient::new(
```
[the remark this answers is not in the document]
jordens: In my attempts it wasn't able.
Maybe that's a future thing.

src/array.rs
jordens: I don't see a way to make it automatic.
The error message however is very convenient and tells you exactly the value it wants.
So during development I don't expect this to be inconvenient.
The fact that all the fragility due to the iterator state size and the talk about its constraints etc is gone is a major plus though.

ryan-summers: Ah, so if you do the wrong value, the compiler should guide you towards the right one? That does sound incredibly convenient. Is that also the case for the `MqttClient` type parameter?

src/mqtt_client/mod.rs
ryan-summers: Appears to not work unfortunately.

tests/integration_test.rs
ryan-summers: Appears to not work unfortunately.

tests/validation_failure.rs
ryan-summers: Appears to not work unfortunately.

src/array.rs
```diff
@@ -106,62 +29,67 @@ const fn digits(x: usize) -> usize {
...
+            }
+
+            fn get_by_key<K, S>(&self, mut keys: K, ser: S) -> Result<usize, Error<S::Error>>
+            where
+                K: Iterator,
+                K::Item: Key,
+                S: serde::Serializer,
+            {
+                let key = keys.next().ok_or(Error::TooShort(0))?;
+                let index = key.find::<$d, Self>().ok_or(Error::NotFound(1))?;
```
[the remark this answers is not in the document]
jordens: Flexibility and stability.
E.g. if you come up with a efficient scheme of packing key indices, you can just implement your unpacker to usize in `find` in your own crate 
Otherwise you'd have to change the trait or even touch each existing impl.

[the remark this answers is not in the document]
jordens: Flexibility and stability.
E.g. if you come up with a efficient scheme of packing key indices, you can just implement your unpacker to usize in `find` in your own crate 
Otherwise you'd have to change the trait or even touch each existing impl.

src/iter.rs
jordens: Without this we don't get the size hint.

src/lib.rs
jordens: We'd have to pull D up into the trait. It doesn't help and would be longer. We always have to explicitly carry the depth anyway.

tests/option.rs
jordens: Defer(0) calls the miniconf API.
No miniconf attribute calls serde directly.
For option, the former is the "mask single serde item" use case.

src/iter.rs
jordens: Metadata calculation isn't free. It's a recursive traversal of the full tree.

src/array.rs
jordens: Iirc yes.

examples/mqtt.rs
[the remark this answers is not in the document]
jordens: Iirc it tells you if you put the wrong number.

tests/arrays.rs
jordens: Right!

miniconf_derive/src/field.rs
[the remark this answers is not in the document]
jordens: Not in my tests.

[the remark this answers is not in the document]
jordens: But we should be able to just `?` the result.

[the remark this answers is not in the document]
ryan-summers: Meh, we can handle this later. It's nbd for now.

ryan-summers (approved):

miniconf_derive/src/field.rs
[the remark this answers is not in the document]
jordens: https://github.com/quartiq/miniconf/pull/172

PR #174, opened 2023-08-02
recursion depth

jordens:
- update the generic bound logic in the derive macro

This rewrites the type bound logic after having analyzed it again and
after the changes due to adding the cont generic in `Miniconf`.

The Miniconf<Y> const generic Y is now the miniconf recursion depth
again. And it's an upper bound to the state size.
Before, the idea of making it exactly the key length war broken:
We had `impl Miniconf<0> for Option<T> where T: SerDe`
but for `#[miniconf(defer(1))] a: [Option<T>; N]` it would not use the
miniconf API of `Option<T>` as desired as it would for
`#[miniconf(defer(0))] a: Option<T>`.
Using the recursion depth as indicator which API to call fixes this.

At the cost of the state size now being potentially a bit too large.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,13 +53,13 @@ struct Settings {
     array_miniconf: [Inner; 2],
 
     // Hiding paths by setting the Option to `None` at runtime
-    #[miniconf(defer(0))]
+    #[miniconf(defer)]
     option_defer: Option<i32>,
     // Hiding a path and deferring to the inner
-    #[miniconf(defer(1))]
+    #[miniconf(defer(2))]
     option_miniconf: Option<Inner>,
     // Hiding elements of an Array of Miniconf items
-    #[miniconf(defer(2))]
+    #[miniconf(defer(3))]
     array_option_miniconf: [Option<Inner>; 2],
 }
 
@@ -141,8 +141,8 @@ Homogeneous [core::array]s can be made accessible either
 
 `Option` is used
 1. like a standard `serde` Option, or
-2. with `#[miniconf(defer(0))]` to support a value that may be absent (masked) at runtime.
-3. with `#[miniconf(defer(D))]` and `D >= 1` to support mask sub-trees at runtime.
+2. with `#[miniconf(defer(1))]` to support a value that may be absent (masked) at runtime.
+3. with `#[miniconf(defer(D))]` and `D >= 2` to support mask sub-trees at runtime.
 
 Structs, arrays, and Options can then be cascaded to construct more complex trees.
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # Miniconf
+
 [![crates.io](https://img.shields.io/crates/v/miniconf.svg)](https://crates.io/crates/miniconf)
 [![docs](https://docs.rs/miniconf/badge.svg)](https://docs.rs/miniconf)
 [![QUARTIQ Matrix Chat](https://img.shields.io/matrix/quartiq:matrix.org)](https://matrix.to/#/#quartiq:matrix.org)
@@ -15,6 +16,7 @@ client](MqttClient) and a Python reference implementation to ineract with it.
 `Miniconf` is completely generic over the `serde::Serializer`/`serde::Deserializer` backend and the path hierarchy separator.
 
 ## Example
+
 ```rust
 use miniconf::{Error, Miniconf, JsonCoreSlash};
 use serde::{Deserialize, Serialize};
@@ -117,6 +119,7 @@ for path in Settings::iter_paths::<String>("/") {
 ```
 
 ## MQTT
+
 There is an [MQTT-based client](MqttClient) that implements settings management over the [MQTT
 protocol](https://mqtt.org) with JSON payloads. A Python reference library is provided that
 interfaces with it.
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -141,8 +141,8 @@ Homogeneous [core::array]s can be made accessible either
 
 `Option` is used
 1. like a standard `serde` Option, or
-2. with `#[miniconf(defer(0))]` to support a value that may be absent (masked) at runtime.
-3. with `#[miniconf(defer(D))]` and `D >= 1` to support mask sub-trees at runtime.
+2. with `#[miniconf(defer(1))]` to support a value that may be absent (masked) at runtime.
```
ryan-summers, line 144: Are `defer` and `defer(1)` equivalent?

ryan-summers: Looks like yes, can we note that?

src/lib.rs
```diff
@@ -149,7 +149,23 @@ impl Key for &str {
...
 /// Trait exposing serialization/deserialization of nodes by keys/paths and traversal.
 ///
-/// The depth parameter `Y` is the maximum key length: the depth/height of the tree.
+/// The depth parameter `Y` is the maximum miniconf recursion depth.
+/// An implementor chooses a `Z <= Y` and may then in the recursive methods (directly
+/// and indirectly) retrieve at most `Z` keys from the `keys` and `indices` iterators
+/// and call the `func` callback at most `Z` times. It may
+/// use at most `Miniconf<Z>` on its inner types. `Z` must be consistent, i.e. if
+/// an implementor retrieves `X` keys before recursing to an inner type, it may at
+/// most use `Miniconf<Y - X>` on its inner types.
```
ryan-summers, line 158: I have absolutely no idea what this means... Is there a  way we can state this in simpler terms?

src/lib.rs
```diff
@@ -149,7 +149,23 @@ impl Key for &str {
...
+///
+/// Furthermore the `derive(Miniconf)` macro assumes that any implementor of `Miniconf<Y>`
+/// will require its generic types to be exactly `Miniconf<Y - 1>`. This condition is upheld
+/// by all implementations in this crate. It is only violated
+/// when deriving `Miniconf` for a struct that (a) forwards its own type parameters as type
+/// parameters to its field types, (b) uses `Miniconf` on those fields, and (c) those field
+/// types use their type parameters at other levels than `Miniconf<Y - 1>`.
```
ryan-summers, line 165: I don't particularly understand why this matters...? Is this something that needs to be upheld, or is this purely informational about the design? Maybe we can move it under a `Design` subheading or something?

tests/generics.rs
```diff
@@ -86,6 +90,17 @@ fn generic_atomic() {
...
+    assert_eq!(metadata.max_length, "/opt1/0/0".len());
+}
+
+#[test]
+fn test_failure() {
+    type A<T> = [[T; 0]; 0];
+    #[derive(Miniconf)]
+    struct S<T: serde::Serialize + serde::de::DeserializeOwned>(
+        // this wrongly infers T: Miniconf<1> instead of T: SerDe
+        // adding the missing bound is a workaround
```
ryan-summers, line 103: What is this actually testing?

tests/option.rs
```diff
@@ -155,3 +155,16 @@ fn option_absent() {
...
+fn array_option() {
+    #[derive(Copy, Clone, Default, Miniconf)]
+    struct S {
+        #[miniconf(defer(1))]
+        a: Option<u32>,
+        #[miniconf(defer(1))]
+        b: [Option<u32>; 1],
+        #[miniconf(defer(2))]
+        c: [Option<u32>; 1],
+    }
```
ryan-summers, line 169: I don't see any actual tests here

src/lib.rs
```diff
@@ -149,7 +149,23 @@ impl Key for &str {
 
 /// Trait exposing serialization/deserialization of nodes by keys/paths and traversal.
 ///
-/// The depth parameter `Y` is the maximum key length: the depth/height of the tree.
+/// The depth parameter `Y` is the maximum miniconf recursion depth.
+/// An implementor chooses a `Z <= Y` and may then in the recursive methods (directly
+/// and indirectly) retrieve at most `Z` keys from the `keys` and `indices` iterators
+/// and call the `func` callback at most `Z` times. It may
```
ryan-summers, line 155: I don't understand these docs - what `keys` and `indices` are they referring to? Seems like `indices` is referring to the member on this trait, but `keys` might be an arg provided? In general, I'm having a hard time understanding these docs and what they're supposed to mean

ryan-summers (changes requested):
I'm very thoroughly lost on what the depth parameters are supposed to be

tests/generics.rs
jordens: Added comments.

tests/option.rs
jordens: Added comment.

README.md
jordens: Added docs.

src/lib.rs
jordens: Clarified the docs. Please check again.

src/lib.rs
jordens: See above. If you understand, let me know if you see a simpler way.

src/lib.rs
jordens: It matters and it's a limitation (always was). See the tests and the added comments there.
I'll add a link from the readme here.

jordens:
> I'm very thoroughly lost on what the depth parameters are supposed to be

What specific questions do you have that are not answered by the docs?

src/lib.rs
```diff
@@ -149,23 +149,92 @@ impl Key for &str {
...
+/// 1. as a leaf like a standard `serde` Option, or
+/// 2. with `#[miniconf(defer(1))]` to support a leaf value that may be absent (masked) at runtime.
+/// 3. with `#[miniconf(defer(D))]` and `D >= 2` to support masking sub-trees at runtime.
+///
+/// ## Bounds on generics
+///
+/// The macro adds bounds to generic types of the struct it is acting on.
+/// E.g. If a generic type parameter `T` of the struct `S<T>`is used as a type parameter to a
+/// field type `a: F1<F2<T>>` the type `T` will be considered to reside at depth `X = 2` (as it is
+/// within `F2` which is within `F1`) and the following bounds will applied:
```
ryan-summers, line 206:
```suggestion
/// within `F2` which is within `F1`) and the following bounds will be applied:
```

src/lib.rs
```diff
@@ -149,23 +149,92 @@ impl Key for &str {
...
+///     #[miniconf(defer)]
+///     data: [u32; 2],
+/// }
+/// #[derive(Miniconf)]
+/// struct Settings {
+///     // Accessed with path `nested/data/0` or `nested/data/1`
+///     #[miniconf(defer(2))]
+///     nested: Nested,
+///
+///     // Accessed with path `external`
```
ryan-summers, line 234:
```suggestion
///     // Accessed with path `/nested/data/0` or `/nested/data/1`
///     #[miniconf(defer(2))]
///     nested: Nested,
///
///     // Accessed with path `/external`
```

Don't we enforce leading prefixes now?

src/lib.rs
```diff
@@ -149,23 +149,92 @@ impl Key for &str {
...
+/// `Option` is used
+/// 1. as a leaf like a standard `serde` Option, or
+/// 2. with `#[miniconf(defer(1))]` to support a leaf value that may be absent (masked) at runtime.
+/// 3. with `#[miniconf(defer(D))]` and `D >= 2` to support masking sub-trees at runtime.
+///
+/// ## Bounds on generics
+///
+/// The macro adds bounds to generic types of the struct it is acting on.
+/// E.g. If a generic type parameter `T` of the struct `S<T>`is used as a type parameter to a
+/// field type `a: F1<F2<T>>` the type `T` will be considered to reside at depth `X = 2` (as it is
```
ryan-summers, line 205: Can we just write this out as:
```rs
type F1<T> = ...;
type F2<T> = ...;

struct S<T> {
    a: F1<F2<T>>,
}
```

It actually took me a couple read throughs to realize this is what it was trying to describe.

src/lib.rs
```diff
@@ -149,23 +149,92 @@ impl Key for &str {
...
 /// Trait exposing serialization/deserialization of nodes by keys/paths and traversal.
 ///
-/// The depth parameter `Y` is the maximum miniconf recursion depth.
-/// An implementor chooses a `Z <= Y` and may then in the recursive methods (directly
-/// and indirectly) retrieve at most `Z` keys from the `keys` and `indices` iterators
-/// and call the `func` callback at most `Z` times. It may
-/// use at most `Miniconf<Z>` on its inner types. `Z` must be consistent, i.e. if
-/// an implementor retrieves `X` keys before recursing to an inner type, it may at
-/// most use `Miniconf<Y - X>` on its inner types.
+/// The keys used to locate nodes can bei either iterators over `usize` or iterators
```
ryan-summers, line 152:
```suggestion
/// The keys used to locate nodes can be either iterators over `usize` or iterators
```

ryan-summers (changes requested):
The docs look much more understandable, thanks for taking the time to rework them

src/lib.rs
jordens: Everything before the first / is ignored. This only applies in json_core
Added doc about that there.

src/lib.rs
jordens: Added doctest.

src/lib.rs
jordens: Ack

ryan-summers:
@jordens Once you update this, we can probably merge

src/lib.rs
```diff
@@ -377,6 +486,16 @@ pub trait Miniconf<const Y: usize = 1> {
     ///
     /// See also [Miniconf::iter_paths].
     ///
+    /// ```
+    /// # use miniconf::{Miniconf, Metadata};
+    /// # use heapless::String;
+    /// #[derive(Miniconf)]
+    /// struct S {foo: u32};
```
ryan-summers, line 503: Gotta remove teh `heapless::String` throughout or fix the import

jordens: It's both needed and correct. The problem is elsewhere.

ryan-summers (approved):

PR #178, opened 2023-08-21
deserialize borrowed

jordens:
- pull in serde/std with std
- support borrowed deserialize

### DIFF
--- a/miniconf_derive/src/lib.rs
+++ b/miniconf_derive/src/lib.rs
@@ -257,7 +257,9 @@ pub fn derive_tree_deserialize(input: TokenStream) -> TokenStream {
     let lts: Vec<_> = input.generics.lifetimes().cloned().collect();
     input.generics.params.push(parse_quote!('de));
     if let Some(syn::GenericParam::Lifetime(de)) = input.generics.params.last_mut() {
+        assert_eq!(de.lifetime.ident, "de");
         for l in lts {
+            assert!(l.lifetime.ident != "de");
             de.bounds.push(l.lifetime);
         }
     }
--- a/miniconf_derive/src/lib.rs
+++ b/miniconf_derive/src/lib.rs
@@ -258,7 +258,7 @@ pub fn derive_tree_deserialize(input: TokenStream) -> TokenStream {
     input.generics.params.push(parse_quote!('de));
     if let Some(syn::GenericParam::Lifetime(de)) = input.generics.params.last_mut() {
         for l in lts {
-            de.bounds.push(l.lifetime.clone());
+            de.bounds.push(l.lifetime);
         }
     }
     let (impl_generics, _, _) = input.generics.split_for_impl();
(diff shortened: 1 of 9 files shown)


miniconf_derive/src/lib.rs
```diff
@@ -252,11 +252,22 @@ pub fn derive_tree_deserialize(input: TokenStream) -> TokenStream {
...
 
-    let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
+    let orig_generics = input.generics.clone();
+    let (_, ty_generics, where_clause) = orig_generics.split_for_impl();
+    let lts: Vec<_> = input.generics.lifetimes().cloned().collect();
+    input.generics.params.push(parse_quote!('de));
+    if let Some(syn::GenericParam::Lifetime(de)) = input.generics.params.last_mut() {
+        assert_eq!(de.lifetime.ident, "de");
+        for l in lts {
+            assert!(l.lifetime.ident != "de");
```
ryan-summers, line 262: We could technically loop through randomized lifetime names until we find one that is unused, but tbh I like just using `'de`. I think it's fairly unlikely that someone will use this in their type, and if so, we catch it here.

miniconf_derive/src/lib.rs
```diff
@@ -252,11 +252,22 @@ pub fn derive_tree_deserialize(input: TokenStream) -> TokenStream {
...
+            de.bounds.push(l.lifetime);
+        }
+    }
+    let (impl_generics, _, _) = input.generics.split_for_impl();
 
     quote! {
-        impl #impl_generics ::miniconf::TreeDeserialize<#depth> for #ident #ty_generics #where_clause {
-            fn deserialize_by_key<'de, K, D>(&mut self, mut keys: K, de: D) -> Result<usize, ::miniconf::Error<D::Error>>
+        impl #impl_generics ::miniconf::TreeDeserialize<'de, #depth> for #ident #ty_generics #where_clause {
```
ryan-summers, line 269: What does this look like for a user that doesn't support borrowed deserialization? Does the lifetime go to anonymous, or is it just `'static`?

jordens: A struct that owns all its data doesn't have lifetimes. It's then just `impl<'de> TreeDeserialize<'de> for S {...}`.
A `Deserializer` backend that doesn't support borrowed data runs into the usual `unimplemented!()` panics (can't be avoided).
The prime example of a user that doesn't support structures with borrowed data is the mqtt client. The bound there is `where for<'de> Settings: JsonCoreSlash<'de, Y> + Clone`, which is analogous to [`DeserializeOwned`](https://docs.rs/serde/latest/src/serde/de/mod.rs.html#609-610).

https://github.com/quartiq/miniconf/pull/178/files#diff-8ce4ae904c14212d6bed3296e5ced8732145b011aa19e6cb2fa5c614698d6f4cR180

ryan-summers: Ah that makes sense. Kind of an odd syntax for the lifetime but whatever.

ryan-summers (approved):

miniconf_derive/src/lib.rs
jordens: We could add `TreeDeserializeOwned` and `JsonCoreSlashOwned` as shorthand traits along with blanket impls. But I think that's not so urgent now and also not much shorter.

PR #179, opened 2023-08-23
Refactoring MQTT client for latest Minimq

ryan-summers:
Follow-on to https://github.com/quartiq/minimq/pull/144

This PR updates the MQTT client to the latest version of Minimq.

### DIFF
--- a/examples/mqtt.rs
+++ b/examples/mqtt.rs
@@ -23,15 +23,18 @@ struct Settings {
 async fn mqtt_client() {
     // Construct a Minimq client to the broker for publishing requests.
     let mut buffer = [0u8; 1024];
-    let localhost: minimq::embedded_nal::IpAddr = "127.0.0.1".parse().unwrap();
-    let mut mqtt: minimq::Minimq<'_, _, _, minimq::broker::IpBroker> = minimq::Minimq::new(
-        Stack::default(),
-        StandardClock::default(),
-        minimq::ConfigBuilder::new(localhost.into(), &mut buffer)
+    let mut mqtt: minimq::Minimq<'_, _, _, minimq::broker::NamedBroker<Stack>> =
+        minimq::Minimq::new(
+            Stack::default(),
+            StandardClock::default(),
+            minimq::ConfigBuilder::new(
+                minimq::broker::NamedBroker::new("localhost", Stack).unwrap(),
+                &mut buffer,
+            )
             .client_id("tester")
             .unwrap()
             .keepalive_interval(60),
-    );
+        );
 
     // Wait for the broker connection
     while !mqtt.client().is_connected() {
@@ -39,6 +42,8 @@ async fn mqtt_client() {
             .unwrap();
         tokio::time::sleep(Duration::from_millis(100)).await;
     }
+
+    log::info!("Test client connected");
 
     // Wait momentarily for the other client to connect.
     tokio::time::sleep(Duration::from_secs(1)).await;
@@ -88,7 +93,7 @@ async fn main() {
     let mut client: miniconf::MqttClient<'_, _, _, _, minimq::broker::IpBroker, 2> =
         miniconf::MqttClient::new(
             Stack::default(),
-            "republish/device",
+            "sample/prefix",
             StandardClock::default(),
             Settings::default(),
             minimq::ConfigBuilder::new(localhost.into(), &mut buffer).keepalive_interval(60),
--- a/examples/mqtt.rs
+++ b/examples/mqtt.rs
@@ -1,4 +1,4 @@
-use miniconf::{MqttClient, Tree};
+use miniconf::Tree;
 use minimq::Publication;
 use std::time::Duration;
 use std_embedded_nal::Stack;
--- a/examples/mqtt.rs
+++ b/examples/mqtt.rs
@@ -1,5 +1,5 @@
 use miniconf::{MqttClient, Tree};
-use minimq::{Minimq, Publication};
+use minimq::Publication;
 use std::time::Duration;
 use std_embedded_nal::Stack;
 use std_embedded_time::StandardClock;
@@ -22,13 +22,19 @@ struct Settings {
 
 async fn mqtt_client() {
     // Construct a Minimq client to the broker for publishing requests.
-    let mut mqtt: Minimq<_, _, 256, 1> = Minimq::new(
-        "127.0.0.1".parse().unwrap(),
-        "tester",
+    let mut rx_buffer = [0u8; 256];
+    let mut tx_buffer = [0u8; 256];
+    let mut session = [0u8; 256];
+    let localhost: minimq::embedded_nal::IpAddr = "127.0.0.1".parse().unwrap();
+    let mut mqtt: minimq::Minimq<'_, _, _, minimq::broker::IpBroker> = minimq::Minimq::new(
         Stack::default(),
         StandardClock::default(),
-    )
-    .unwrap();
+        minimq::Config::new(localhost.into(), &mut rx_buffer, &mut tx_buffer)
+            .client_id("tester")
+            .unwrap()
+            .session_state(&mut session)
+            .keepalive_interval(60),
+    );
 
     // Wait for the broker connection
     while !mqtt.client().is_connected() {
@@ -78,15 +84,25 @@ async fn main() {
     // Spawn a task to send MQTT messages.
     tokio::task::spawn(async move { mqtt_client().await });
 
-    let mut client: MqttClient<_, _, _, 256, 2> = MqttClient::new(
-        Stack::default(),
-        "",
-        "sample/prefix",
-        "127.0.0.1".parse().unwrap(),
-        StandardClock::default(),
-        Settings::default(),
-    )
-    .unwrap();
+    let mut rx_buffer = [0u8; 256];
+    let mut tx_buffer = [0u8; 256];
+    let mut session = [0u8; 256];
+    let localhost: minimq::embedded_nal::IpAddr = "127.0.0.1".parse().unwrap();
+
+    // Construct a settings configuration interface.
+    let mut client: miniconf::MqttClient<'_, _, _, _, minimq::broker::IpBroker, 2> =
+        miniconf::MqttClient::new(
+            Stack::default(),
+            "republish/device",
+            StandardClock::default(),
+            Settings::default(),
+            minimq::Config::new(localhost.into(), &mut rx_buffer, &mut tx_buffer)
+                .client_id("tester")
+                .unwrap()
+                .session_state(&mut session)
+                .keepalive_interval(60),
+        )
+        .unwrap();
 
     loop {
         if client.update().unwrap() {
--- a/examples/mqtt.rs
+++ b/examples/mqtt.rs
@@ -22,17 +22,14 @@ struct Settings {
 
 async fn mqtt_client() {
     // Construct a Minimq client to the broker for publishing requests.
-    let mut rx_buffer = [0u8; 256];
-    let mut tx_buffer = [0u8; 256];
-    let mut session = [0u8; 256];
+    let mut buffer = [0u8; 1024];
     let localhost: minimq::embedded_nal::IpAddr = "127.0.0.1".parse().unwrap();
     let mut mqtt: minimq::Minimq<'_, _, _, minimq::broker::IpBroker> = minimq::Minimq::new(
         Stack::default(),
         StandardClock::default(),
-        minimq::Config::new(localhost.into(), &mut rx_buffer, &mut tx_buffer)
+        minimq::ConfigBuilder::new(localhost.into(), &mut buffer)
             .client_id("tester")
             .unwrap()
-            .session_state(&mut session)
             .keepalive_interval(60),
     );
 
@@ -84,9 +81,7 @@ async fn main() {
     // Spawn a task to send MQTT messages.
     tokio::task::spawn(async move { mqtt_client().await });
 
-    let mut rx_buffer = [0u8; 256];
-    let mut tx_buffer = [0u8; 256];
-    let mut session = [0u8; 256];
+    let mut buffer = [0u8; 1024];
     let localhost: minimq::embedded_nal::IpAddr = "127.0.0.1".parse().unwrap();
 
     // Construct a settings configuration interface.
@@ -96,11 +91,7 @@ async fn main() {
             "republish/device",
             StandardClock::default(),
             Settings::default(),
-            minimq::Config::new(localhost.into(), &mut rx_buffer, &mut tx_buffer)
-                .client_id("tester")
-                .unwrap()
-                .session_state(&mut session)
-                .keepalive_interval(60),
+            minimq::ConfigBuilder::new(localhost.into(), &mut buffer).keepalive_interval(60),
         )
         .unwrap();
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,6 +26,9 @@ mod mqtt_client;
 #[cfg(feature = "mqtt-client")]
 pub use mqtt_client::*;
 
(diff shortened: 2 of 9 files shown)


src/mqtt_client.rs
```diff
@@ -290,12 +281,10 @@ where
...
 
-            // Note: The topic may be absent at runtime (deferred `Option`).
-            let len = match self.settings.get_json(&topic, &mut data) {
-                Err(Error::Absent(_)) => continue,
-                Ok(len) => len,
-                e => e.unwrap(),
-            };
+            let payload = minimq::publication::DeferedPayload::new(|buf| {
+                // TODO: Figure out how to handle absent topics.
+                Ok(self.settings.get_json(&topic, buf).unwrap())
```
ryan-summers, line 286: @jordens My current thought on this is that we update the `republish_state()` iterator to allow us to simply skip absent topics. Would make this simple, so we can selectively iterate over only the present settings if we want.

src/mqtt_client.rs
```diff
@@ -484,6 +473,7 @@ where
                             // Note(unwrap): The vector is guaranteed to be as large as the largest MQTT
                             // message size, so the properties (which are a portion of the message) will
                             // always fit into it.
+                            // TODO: Figure out how to handle cached properties for listing?
```
ryan-summers, line 476: Current thought is that we just don't cache the properties anymore. Instead, we could only cache:
* The response topic
* The correlation data

Still not entirely sure how to properly cache this data, and I'm not sure if it's required. Perhaps we could just entirely omit the properties when republishing. Alternatively, the user could provide yet another buffer to miniconf to store this cached data into?

src/mqtt_client.rs
```diff
@@ -500,35 +490,38 @@ where
...
-
-                            if client.publish(message).is_err() {
-                                Response::error("Can't publish `Get` response")
-                            } else {
-                                Response::ok()
+                    let payload = minimq::publication::DeferedPayload::new(|buf| {
+                        match self.settings.get_json(path, buf) {
+                            Ok(len) => len,
+                            Err(err) => {
+                                // TODO: Lookup failure. Serialize the error as a response.
```
ryan-summers, line 497: I think the best thing to do here is to do a proactive check on the `path` before attempting to defer the payload. If the path doesn't exist, then we bail out early. Open to thoughts on this one

src/mqtt_client.rs
```diff
@@ -567,6 +561,7 @@ where
                     // always fit into it.
                     self.properties_cache
                         .replace(Vec::from_slice(binary_props).unwrap());
+                    // TODO: Figure out how to handle cached responses
```
ryan-summers, line 564: I think the only time we realistically will cache a response is when we receive a `Get` command. Specifcally because when we receive this command, we do two things:
1. Publish the `path` requested by the `Get` command
2. Reply to the `Get` command

I imagine we could simplify this into a single response somehow, where it's either:
1. Get command succeeded, topic posted to the proper path
2. Get command failed, response provided to the ResponseTopic

With this structure, all requests get exactly 1 reply, which means we don't need to bother with caching anymore.

ryan-summers:
@jordens I'm fighting a few issues currently and wanted to post them here so you could comment on them. I'm curious to hear some of your thoughts. I suspect we may need to add some new `Miniconf` APIs to address some of these changes

src/mqtt_client.rs
ryan-summers: The difficulty arises because with the current architecture, we won't realize the path doesn't exist until we go to serialize the publish packet for transmission. I don't know if we can propogate the error all the way back through the stack at that point, so we lose the exact error intel. It could have failed serialization due to a number of other problems, like:
1. No response topic identified
2. Not enough buffer space

src/mqtt_client.rs
jordens: The iterator doesn't know about absent topics. We'd have to extend/change the trait.

What about just erroring from that callback and failing the packet serialization?

src/mqtt_client.rs
jordens: Don't we always send exactly one packet? I.e.

* start serializing the response packet
* as part of that, try the operation (in the callback, late)
* fail operation for reason Err (absent, not found, buffer too small...) or succeed wit Ok
* try serializing the Err/Ok into the response packet
* if that fails, serialize whatever fixed fatal error message can definitely be serialized
* send the packet

src/mqtt_client.rs
jordens: Agreed. That's better. But do send the correlation data both cases.

src/mqtt_client.rs
jordens: ;) Well no, we don't. See [below](https://github.com/quartiq/miniconf/pull/179/files#r1302893601) for that discussion.

jordens: The question is then can we change the topic (from the get topic to the response topic) while serializing if we hit an err?

src/mqtt_client.rs
jordens: The other option would be to always send the get response to the response topic (Ok or Err) and extend the list command to also (or invent a new command) to be an actual re-publication run (as on startup).

src/mqtt_client.rs
ryan-summers: I have opted to drop the caching of properties, so CD isn't maintained through all the listed elements - just the initial reply to the list request.

src/mqtt_client.rs
ryan-summers: I figured out how to handle this. I updated minimq to allow the error to propagate back from the `publish()` call, so if the serialization fails, we can handle it and retain the error type.

src/mqtt_client.rs
ryan-summers: I updated it to properly handle this condition now.

src/mqtt_client.rs
ryan-summers: Opted for:
1. The `Get` command only sends one reply (always has CD) to the response topic.
2. Failure to serialize in the `Get` command (i.e. path is invalid) will result in a generic failure being sent

src/mqtt_client.rs
```diff
@@ -159,64 +159,57 @@ impl<'a> Command<'a> {
...
         let mut connection_topic = prefix.clone();
         connection_topic.push_str("/alive").unwrap();
-        mqtt.client().set_will(
-            &connection_topic,
-            b"0",
-            QoS::AtMostOnce,
-            Retain::Retained,
-            &[],
-        )?;
+        let mut will = minimq::Will::new(&connection_topic, b"0", &[]).unwrap();
```
ryan-summers, line 203: Gotta figure out the borrowing of the will topic here... The topic doesn't live long enough to just be borrowed

src/mqtt_client.rs
```diff
@@ -159,64 +166,58 @@ impl<'a> Command<'a> {
...
+        will.retained(Retain::Retained);
+        will.qos(QoS::AtMostOnce);
+
+        let mut config = config
+            .keepalive_interval(KEEPALIVE_INTERVAL_SECONDS)
+            .autodowngrade_qos();
+
+        config.will(will).ok();
```
ryan-summers, line 218: Not sure what the best choice of action here is if the user doesn't provide us with `will_buffer` memory in the MQTT client config.

ryan-summers: Current plan is to refactor the will buffer management.

examples/mqtt.rs
```diff
@@ -22,13 +22,16 @@ struct Settings {
...
-        "127.0.0.1".parse().unwrap(),
-        "tester",
+    let mut buffer = [0u8; 1024];
+    let localhost: minimq::embedded_nal::IpAddr = "127.0.0.1".parse().unwrap();
+    let mut mqtt: minimq::Minimq<'_, _, _, minimq::broker::IpBroker> = minimq::Minimq::new(
         Stack::default(),
         StandardClock::default(),
-    )
-    .unwrap();
+        minimq::ConfigBuilder::new(localhost.into(), &mut buffer)
```
jordens, line 30: Why does this take a `&mut [u8, N]` and not a `&mut [u8]`?

examples/mqtt.rs
```diff
@@ -22,13 +22,16 @@ struct Settings {
 
 async fn mqtt_client() {
     // Construct a Minimq client to the broker for publishing requests.
-    let mut mqtt: Minimq<_, _, 256, 1> = Minimq::new(
-        "127.0.0.1".parse().unwrap(),
-        "tester",
+    let mut buffer = [0u8; 1024];
+    let localhost: minimq::embedded_nal::IpAddr = "127.0.0.1".parse().unwrap();
```
jordens, line 26: can we try `localhost` (and thus DNS) here?

src/mqtt_client.rs
```diff
@@ -159,64 +168,58 @@ impl<'a> Command<'a> {
...
-            QoS::AtMostOnce,
-            Retain::Retained,
-            &[],
-        )?;
+        let mut will = minimq::Will::new(&connection_topic, b"0", &[])?;
+        will.retained(Retain::Retained);
+        will.qos(QoS::AtMostOnce);
+
+        let config = config
+            .keepalive_interval(KEEPALIVE_INTERVAL_SECONDS)
```
jordens, line 217: Can we leave this for the user to decide?

src/mqtt_client.rs
```diff
@@ -159,64 +168,58 @@ impl<'a> Command<'a> {
...
+            .autodowngrade_qos();
+
+        let config = config.will(will)?;
```
jordens, line 220:
```suggestion
            .autodowngrade_qos()
            .will(will)?;
```

src/mqtt_client.rs
```diff
@@ -159,64 +168,58 @@ impl<'a> Command<'a> {
...
         connection_topic.push_str("/alive").unwrap();
-        mqtt.client().set_will(
-            &connection_topic,
-            b"0",
-            QoS::AtMostOnce,
-            Retain::Retained,
-            &[],
-        )?;
+        let mut will = minimq::Will::new(&connection_topic, b"0", &[])?;
+        will.retained(Retain::Retained);
```
jordens, line 213: We should probably either have named them `set_retained()` and `set_qos()` or make them a builder pattern (`let will = Will::new(..).retained(..).qos(..);`)

src/mqtt_client.rs
```diff
@@ -305,15 +301,18 @@ where
...
+                // If the topic is not present, we'll fail to serialize the setting into the
+                // payload and will never publish. The iterator has already incremented, so this is
+                // acceptable.
+                DeferredPublication::new(|buf| self.settings.get_json(&topic, buf))
+                    .topic(&prefixed_topic)
+                    .finish()
+                    .unwrap(),
+            ) {
+                Err(minimq::PubError::Serialization(Error::Absent(_))) => {}
+                other => other.unwrap(),
```
jordens, line 314: Should we not catch `miniconf::Error::Inner()`, i.e. buf too short instead of panicing?

src/mqtt_client.rs
```diff
@@ -480,55 +432,45 @@ where
...
+                            .finish()
+                    else {
+                        return;
+                    };
+
+                    match client.publish(message) {
+                        Err(minimq::PubError::Serialization(error)) => error.into(),
+
+                        // Otherwise, we should consider the response sent.
+                        _ => return,
```
jordens, line 473: Could you comment what this returns from? I can't keep the return/continue/break behavior w.r.t match/closures/loops in my head.

tests/integration_test.rs
```diff
@@ -88,27 +88,29 @@ impl Timer {
...
+    let localhost: minimq::embedded_nal::IpAddr = "127.0.0.1".parse().unwrap();
+
+    // Construct a settings configuration interface.
+    let mut interface: miniconf::MqttClient<'_, _, _, _, minimq::broker::IpBroker, 2> =
+        miniconf::MqttClient::new(
+            Stack,
+            "device",
+            StandardClock::default(),
+            Settings::default(),
+            minimq::ConfigBuilder::new(localhost.into(), &mut buffer).keepalive_interval(60),
```
jordens, line 111: Ah. Doesn't this clash with the const keepalive in `new()`?

jordens (changes requested):

src/lib.rs
```diff
@@ -26,6 +26,9 @@ mod mqtt_client;
 #[cfg(feature = "mqtt-client")]
 pub use mqtt_client::*;
 
+#[cfg(feature = "mqtt-client")]
+pub use minimq;
```
jordens, line 30: Do we really need this?
Or can we make the part of minimq that miniconf exposes so flexible that using it doesn't require access to additional minimq api to use it? Configbuilder is one of the main blockers here, right?

ref https://github.com/rust-lang/api-guidelines/discussions/176 for some discussion.

examples/mqtt.rs
ryan-summers: I'm not sure what you're asking? Do you want to downconvert the buffer using i.e. `&mut buffer[..]`? Not sure what the benefit would be

tests/integration_test.rs
ryan-summers: Yeah technically, but they're all 60s. It hink this was just copy-pasta.

src/lib.rs
ryan-summers: The minimal API surface would be the `minimq::ConfigBuilder` and all the types that touch it (i.e. buffer config enum, `ProtocolError`, and `Broker` types).

That discussion seems to explicitly recommend a `pub use minimq` here, so I'm a bit confused

src/mqtt_client.rs
ryan-summers: It's not clear to me what handling that issue would actually do. I don't think we want to ignore those issues and continue operating, as the user of miniconf would then silently have potential settings that just aren't getting published and/or cannot be set.

What would we do instead of panicking here?

examples/mqtt.rs
jordens: Ah. Here it coerces.
https://github.com/quartiq/minimq/blob/2b466aa1bd7377d59ba58f76172bdd3422d5570c/src/config.rs#L50

src/lib.rs
jordens: I think I pasted the wrong link. I has read somewhere else (and that sounds very reasonable to me) that because of possible semver hazards, the re-exports should be absolutely minimal to allow meeting type requirements of the re-exporting crate but to minimize knock-on breaking changes due unrelated stuff changing in the re-exported crate.
Strictly we should only be re-exporting configbuilder, broker, and the relevant error type.
But that's already a lot, hence let's stick with a full re-export.

src/mqtt_client.rs
ryan-summers: If we handle this message and log it, I have concerns. Specifically:
1. Logging is not required on the end-user side, so they may be unaware of the issue
2. A silently unavailable settings seems very problematic

I honestly think an internal panic here is appropriate because it immediately informs the developer at the development stage to fix the issue and doesn't allow it to potentially seep into a production build.

jordens: I agree that for development a panic is very desirable here.
But for production, the panic is worse than most other options, especially if we know what we can do.
What about instead sending a message to that republication topic, either empty, with some error property attached, or with "<serialization failure>" as payload? Then there is feedback that something is wrong.

ryan-summers: I think those are both good ideas, I'll take a look at what makes the most sense.

ryan-summers:
@jordens Do we want to wait until minimq releases to land this, or are we fine with leaving a `master` branch dependency until we re-release miniconf?

jordens (approved):
Looks good!
Let's merge like this, try on stabilizer etc and then do the release ripples.

PR #183, opened 2023-10-12
Adding support for a `tree(skip)` field attribute

ryan-summers:
Adds support for skipping fields in a miniconf tree.

Fixes #182

PR #185, opened 2023-11-01
Adding unit tests for skipped attributes

ryan-summers:
I forgot to check in these unit tests alongside https://github.com/quartiq/miniconf/pull/183. Also fixed a clippy lint.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,11 +10,9 @@ pub use miniconf_derive::*;
 mod tree;
 pub use tree::*;
 mod array;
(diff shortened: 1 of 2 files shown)


jordens (commented):
we don't see the clippy lint because it's only in nightly.
do we want to use clippy nightly in CI?
Otherwise LGTM

ryan-summers:
Meh, I don't see any reason to personally. It's not a major lint, and lints aren't deal breakers to me. We'll pick them up as Rust releases

PR #189, opened 2023-12-04
add indices iterator

### LINKED ISSUE #188: add `TreeKey::iter_indices{,_unchecked}()`

jordens:
close #188


src/iter.rs
```diff
@@ -113,3 +113,103 @@ where
...
+
+impl<M, const Y: usize> Iterator for IndexIter<M, Y>
+where
+    M: TreeKey<Y> + ?Sized,
+{
+    type Item = ([usize; Y], usize);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            return match M::traverse_by_key(self.state.iter().copied(), |_, _| Ok::<(), ()>(())) {
```
ryan-summers, line 168: DO we really have to copy the iterator? That seems odd to me.

src/iter.rs
```diff
@@ -113,3 +113,103 @@ where
...
+                    if self.count == Some(0) {
+                        None
+                    } else {
+                        debug_assert_eq!(self.count.unwrap_or(1), 1);
+                        self.count = Some(0);
+                        Some((self.state, 1))
+                    }
+                }
+                // Non-root leaf: advance index at current depth
+                Ok(depth) => {
```
ryan-summers, line 194:
```suggestion
                Ok(depth @ 1..) => {
```

Maybe to future proof it a little better?

tests/iter.rs
```diff
@@ -25,6 +25,16 @@ fn struct_iter() {
     assert_eq!(paths.next(), None);
 }
 
+#[test]
+fn struct_iter_indices() {
+    let mut paths = [([0, 0], 1), ([1, 0], 1), ([2, 0], 2)].into_iter();
+    for (have, expect) in Settings::iter_indices().zip(&mut paths) {
```
ryan-summers, line 31: It would be a good sanity to check to also add an access of the item via the iterated indices and verify they point to what we think they do.

ryan-summers: Maybe convert the indices to a path for each and verify they are what we think they should be.

tests/iter.rs
```diff
@@ -25,6 +25,16 @@ fn struct_iter() {
...
+        assert_eq!(have, expect);
+    }
+    // Ensure that all fields were iterated.
+    assert_eq!(paths.next(), None);
+}
+
```
ryan-summers, line 37: It would be good to have a test case for the special "single root option type" thats covered by the `Ok(0)` branch in the iter methods.

ryan-summers (changes requested):
Looking good! Few comments below.

Mind:
- [ ] Adding a CHANGELOG entry

src/iter.rs
jordens: That's not what this is. It copies elements.

ryan-summers: Ah gotcha, so you generate `T::copy()` instead of `&T`, gotcha. Thanks

tests/iter.rs
jordens: https://github.com/quartiq/miniconf/blob/bd88587de0ca5fc92724122f5b91f4cbdf22f35d/tests/option.rs#L17

ryan-summers (approved):

PR #191, opened 2024-01-18
deps, heapless 0.8

jordens:
- make serde derive optional
- heapless 0.8

### DIFF
--- a/src/mqtt_client.rs
+++ b/src/mqtt_client.rs
@@ -629,13 +629,8 @@ fn handle_listing_request(
         None
     };
 
-    if response_topic.len() > MAX_TOPIC_LENGTH {
-        return Err("Response topic too long");
-    }
-
     Ok(ListCache {
-        // `TryFrom` is misleading as it just calls `Into`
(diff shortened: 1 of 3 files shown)


src/mqtt_client.rs
```diff
@@ -629,13 +629,8 @@ fn handle_listing_request(
         None
     };
 
-    if response_topic.len() > MAX_TOPIC_LENGTH {
-        return Err("Response topic too long");
-    }
-
     Ok(ListCache {
-        // `TryFrom` is misleading as it just calls `Into`
-        topic: String::from(response_topic),
```
ryan-summers, line 638: Didn't we just change this?

jordens: Yes. It was actually a bug before. Now it needs to be try from

ryan-summers: Ah sorry when I saw this, I thought we were reverting a change we just made. I looked at the diff backwards

ryan-summers (approved):

ryan-summers (approved):

PR #193, opened 2024-04-03
mqtt: do not own settings

jordens:
in support and exploration of
https://github.com/quartiq/stabilizer/pull/862

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Python lib signatures have changed (Miniconf.create(), discover())
 * Python lib discovery timeout has been optimized to work well for both slow
   connections (high RTT) and fast ones
-* The MQTT client does not own the minitonf settings struct anymore.
+* The MQTT client does not own the miniconf settings struct anymore.
 
 ### Added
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Python lib signatures have changed (Miniconf.create(), discover())
 * Python lib discovery timeout has been optimized to work well for both slow
   connections (high RTT) and fast ones
+* The MQTT client does not own the minitonf settings struct anymore.
```
ryan-summers, line 14:
```suggestion
* The MQTT client does not own the miniconf settings struct anymore.
```

ryan-summers (approved):

ryan-summers:
It's quite neat that we can manage topic listing using solely the type and we don't need the actual settings structure as well.

jordens:
@ryan-summers I made one more change to the validation handler logic to remove copies in hot paths.

One conceptual issue I found is that if validation fails, we don't republish the actual old settings and all that was visible on MQTT is the new (but invalid and inactive) setting. This (the validation handler taking a different setting than the one that flew over MQTT) is the same class of loss-of-sync issue that we have when we change a setting via USB. Unsure what the solution is.

ryan-summers:
> @ryan-summers I made one more change to the validation handler logic to remove copies in hot paths.
> 
> One conceptual issue I found is that if validation fails, we don't republish the actual old settings and all that was visible on MQTT is the new (but invalid and inactive) setting. This (the validation handler taking a different setting than the one that flew over MQTT) is the same class of loss-of-sync issue that we have when we change a setting via USB. Unsure what the solution is.

I believe this was an intentional design decision. The user is supposed to trigger a republication if they reject settings in their handler.

ryan-summers (approved):

jordens:
How would they do that?

jordens:
Ah!

PR #196, opened 2024-04-08
split mqtt

jordens:
- workspace
- move mqtt client into miniconf_mqtt crate
- removed the readback example as that is comprehensively tested in the readme already and was stale


miniconf/src/lib.rs
```diff
@@ -19,11 +19,6 @@ mod json_core;
 #[cfg(feature = "json-core")]
 pub use json_core::*;
 
-#[cfg(feature = "mqtt-client")]
-mod mqtt_client;
-#[cfg(feature = "mqtt-client")]
-pub use mqtt_client::*;
-
 #[cfg(feature = "mqtt-client")]
```
ryan-summers, line 22: Shouldn't `minimq` be removed too?

ryan-summers (changes requested):
Looks good except for one errant re-export

ryan-summers (approved):

PR #198, opened 2024-04-12
validator

jordens:
- use darling, change attribute syntax, default to depth=0
- add support for validation callbacks
- changelog and docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,12 +16,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * The MQTT settings validation handler now receives the already updated new settings
   and the old settings (instead of the not updated current settings and the new settings).
 * The MQTT client has been split into its own `miniconf_mqtt` crate.
+* The attribute syntax has changed from `#[tree(depth(1))]` to `#[tree(depth=1)]`.
+* The default depth is `0`, also in the case where a `#[tree()]` without `depth` has been specified.
 
 ### Added
 
 * Python lib: Support for clearing a retained setting
 * Python CLI: get() support
 * `TreeKey::iter_indices()` and `iter_indices_unchecked()`
+* Support for validation callbacks and errors to check deserialized data.
 
 ## [0.9.0](https://github.com/quartiq/miniconf/compare/v0.8.0...v0.9.0)
 
--- a/miniconf/src/tree.rs
+++ b/miniconf/src/tree.rs
@@ -264,10 +264,10 @@ impl Key for &str {
 /// #[derive(TreeKey)]
 /// struct S {
 ///     // "/b/1/2" = 5
-///     #[tree(depth(2))]
+///     #[tree(depth=2)]
 ///     b: [[u32; 3]; 3],
 ///     // "/c/0" = [3,4], optionally absent at runtime
-///     #[tree(depth(2))]
+///     #[tree(depth=2)]
 ///     c: [Option<[u32; 2]>; 2],
 /// }
 /// ```
@@ -290,7 +290,7 @@ impl Key for &str {
 /// # use miniconf::TreeKey;
 /// #[derive(TreeKey)]
 /// struct S<T> {
-///     #[tree(depth(3))]
+///     #[tree(depth=3)]
 ///     a: [Option<T>; 2],
 /// };
 /// // This works as [u32; N] implements TreeKey<1>:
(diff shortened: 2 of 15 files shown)


miniconf/src/tree.rs
```diff
@@ -302,9 +315,31 @@ impl Key for &str {
...
+/// types use their type parameters at other levels than `TreeKey<Y - 1>`. See also the
 /// `test_derive_macro_bound_failure` test in `tests/generics.rs`.
 ///
+/// ## Validation
+///
+/// The `TreeDeserialize` derive macro supports validation callbacks.
+/// For leaf fields the signature is `fn(struct: &S, new: T, ident: &str, old: &T) -> Result<T, &str>`
+/// The callback returns `Ok(new)` if the value is to be set. If the callback returns
+/// an `Err(&str)`, the update is aborted and the value remains unchanged.
+/// For intermediate fields the signature is `fn(new: &mut T, ident: &str) -> Result<(), &str>`.
```
ryan-summers, line 327: I'm not sure what this is referring to with "intermediate fields". Perhaps we can add an example below to make it clear?

CHANGELOG.md
```diff
@@ -16,12 +16,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * The MQTT settings validation handler now receives the already updated new settings
   and the old settings (instead of the not updated current settings and the new settings).
 * The MQTT client has been split into its own `miniconf_mqtt` crate.
+* The attribute syntax has changed from `#[tree(depth(1))]` to `#[tree(depth=1)]`.
```
ryan-summers, line 19: Need to add a note somewhere that `handled_update` was removed from the MQTT client. Do we want a separate CHANGELOG for that?

CHANGELOG.md
```diff
@@ -16,12 +16,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+* The default depth is `0`, also in the case where a `#[tree()]` without `depth` has been specified.
```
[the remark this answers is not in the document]
jordens: Sure. There's enough pending stuff to do that anyway.

miniconf/tests/validate.rs
```diff
@@ -0,0 +1,45 @@
+use miniconf::{Error, JsonCoreSlash, Tree};
+
+#[derive(Tree, Default)]
+struct Inner {
+    a: f32,
+}
+
+impl Inner {
+    fn check(&mut self, _field: &str) -> Result<(), &'static str> {
+        (self.a >= 0.0).then_some(()).ok_or("Must not be negative.")
```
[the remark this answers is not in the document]
jordens: Can't do it any other way. To do the deserialization, we have to descend all the way. The value is set before returning up the call stack. We could give these intermediate callbacks a different name to disinguish them from leaf callbacks.

CHANGELOG.md
jordens: Same as https://github.com/quartiq/miniconf/pull/198#discussion_r1565416042

miniconf/tests/validate.rs
```diff
@@ -0,0 +1,45 @@
...
+        Err(Error::Invalid(1, _))
+    ));
+    assert_eq!(s.v, 1.0); // remains unchanged
+    s.set_json("/i/a", "1.0".as_bytes()).unwrap();
+    assert_eq!(s.i.a, 1.0);
+    assert!(matches!(
+        s.set_json("/i/a", "-1.0".as_bytes()),
+        Err(Error::Invalid(1, _))
+    ));
+    assert_eq!(s.i.a, -1.0); // changes as validation failed at higher level
```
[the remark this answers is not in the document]
jordens: Same as https://github.com/quartiq/miniconf/pull/198#discussion_r1565416042

miniconf/tests/validate.rs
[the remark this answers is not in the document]
ryan-summers: Ahhh okay that makes sense. Can we add docs somewhere explaining that design decision? It seems important to let the user know that settings may change even if validation failed.

miniconf/src/tree.rs
jordens: We use `non-leaf` elsewhere. Taking that.

ryan-summers: I still think adding an example code block to the docs may be helpful, but yeah, non-leaf would be more in-line with existing docs.

ryan-summers (approved):
Looks like a nice change

PR #199, opened 2024-04-15
setter/getter

jordens:
* more versatile than validation and with more concise/natural semantics, strictly a superset
* also fix mqtt tests that got unintentionally disabled by residual feature flags


ryan-summers (commented):
Ah this indeed does seem like a nicer implementation. Are there any concerns w.r.t the minimq client about s
Republishing an i.e invalid structure?

ryan-summers:
Yeah my main question was how `miniconf-mqtt` would handle it. Does it properly get forwarded up the error chain, or is it unwrapped?

ryan-summers (approved):

PR #200, opened 2024-04-18
pack

jordens:
Four big things and some smaller tweaks in one PR (sorry @ryan-summers)

* The `Packed` keys implementation for very compact node keys
* Revamped the `Keys`/`Iterator<Item=Key>` stuff
* Added `postcard` support, which leads to very small key-value access with `Packed` keys
* Revamped the `Iterator` architecture a bit to reduce code duplication, especially in the high-complexity area

Changes:

- add packed keys
- packed tweaks
- rewrite packed
- split key, packed, rework packed
- introduce generic iterator
- port PathIter
- port IndexIter to generic iter
- add PackedIter
- make generic iter non-generic over treekey
- simplify iter creation
- Keys::next: don't be generic
- packed corner cases
- packed: aligned, docs
- rm len(), simplify iter, docs/changelog
- expand packed
- tweaks
- deref for Packed
- tweak packed
- changelog
- refactor JsonCoreSlash impl
- fix miniconf_mqtt dependency
- postcard: add blanket implementation
- add comment about ExactSizeIterator

### DIFF
--- a/miniconf/src/iter.rs
+++ b/miniconf/src/iter.rs
@@ -9,7 +9,8 @@ struct Iter<const Y: usize> {
     /// The iteration state.
     ///
     /// It contains the current field/element index at each path hierarchy level
-    /// and needs to be at least as large as the maximum path depth.
+    /// and needs to be at least as large as the maximum path depth (ensured by
+    /// `TreeKey<Y>` contract).
     state: [usize; Y],
 
     /// The remaining length of the iterator.
@@ -55,26 +56,26 @@ impl<const Y: usize> Iter<Y> {
             // Found a leaf at the root: leaf Option/newtype
             // Since there is no way to end iteration by hoping for `NotFound` on a leaf Option,
             // we force the count to Some(0) and trigger on that.
+            Ok(0) if matches!(self.count, Some(0)) => State::Done,
             Ok(0) => {
-                if self.count == Some(0) {
-                    State::Done
-                } else {
-                    debug_assert_eq!(self.count.unwrap_or(1), 1);
-                    self.count = Some(0);
-                    State::Leaf(0)
-                }
+                debug_assert_eq!(self.count.unwrap_or(1), 1);
+                self.count = Some(0);
+                State::Leaf(0)
             }
-            // Non-root leaf: advance index at current depth
-            Ok(depth @ 1..) => {
+            // Non-root leaf (depth @ 1..): advance index at current depth
+            Ok(depth) => {
                 self.count = self.count.map(|c| c - 1);
                 self.state[depth - 1] += 1;
                 State::Leaf(depth)
             }
             Err(Error::Inner(e)) => State::Err(e),
-            // * NotFound(0) Not having consumed any name/index, the only possible case
-            //   is a leaf (e.g. `Option` or newtype), those however can not return `NotFound`.
-            // * TooShort is excluded by construction.
-            // * No other errors are returned by traverse_by_key()/path()
+            // Note(unreachable):
+            // * NotFound(0): Not having consumed any name/index, the only possible case
+            //   is a root leaf (e.g. `Option` or newtype), those however can not return
+            //   `NotFound` as they don't do key lookup.
+            // * TooShort: Excluded by construction (`state.len() == Y` and `Y` being an
+            //   upper bound to key length.
+            // * No other errors are returned by traverse_by_key()/path()/packed()
             _ => unreachable!(),
         }
     }
@@ -89,7 +90,6 @@ impl<const Y: usize> Iter<Y> {
 pub struct PathIter<'a, M: ?Sized, const Y: usize, P> {
     iter: Iter<Y>,
     pm: PhantomData<(P, M)>,
-    /// The separator before each name.
     separator: &'a str,
 }
 
@@ -128,7 +128,7 @@ where
                 }
                 State::Leaf(_depth) => Some(Ok(path)),
                 State::Done => None,
-                State::Err(e @ core::fmt::Error) => Some(Err(e)),
+                State::Err(e) => Some(Err(e)),
             };
         }
     }
--- a/miniconf/src/iter.rs
+++ b/miniconf/src/iter.rs
@@ -1,6 +1,8 @@
 use crate::{Error, Packed, TreeKey};
 use core::{fmt::Write, marker::PhantomData};
 
+// core::iter::ExactSizeIterator would be applicable if `count.is_some()`.`
+
 /// An iterator over nodes in a `TreeKey`.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 struct Iter<const Y: usize> {
--- a/miniconf/src/iter.rs
+++ b/miniconf/src/iter.rs
@@ -39,7 +39,7 @@ enum State<E> {
 }
 
 impl<const Y: usize> Iter<Y> {
-    fn next<E>(&mut self, ret: Result<usize, Error<E>>) -> State<E> {
+    fn handle<E>(&mut self, ret: Result<usize, Error<E>>) -> State<E> {
         match ret {
             // Out of valid indices at the root: iteration done
             Err(Error::NotFound(1)) => {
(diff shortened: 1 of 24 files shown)


miniconf/src/packed.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// * `bits`: Number of bits to push. `bits <= Self::CAPACITY`
+    /// * `value`: Value to push. `value >> bits == 0`
+    #[inline]
+    pub fn push_lsb(&mut self, bits: u32, value: usize) -> Option<u32> {
+        debug_assert_eq!(value >> bits, 0);
+        let mut n = self.trailing_zeros();
+        let m = 1 << n;
+        if let Some(marker) = Self::new(m >> bits) {
+            n -= bits;
+            self.0 = (self.get() ^ m) | ((value << n) << 1) | marker.0;
```
ryan-summers, line 176: I think this could use some docs. We're:
1. Removing the old marker (to clear the bit for data)
2. Appending the new bits on after the existing data
3. Setting the new marker

Took me a while to parse all that.

miniconf/src/packed.rs
```diff
@@ -0,0 +1,219 @@
...
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test() {
+        let t = [1usize, 3, 4, 0, 1];
```
ryan-summers, line 208: It would be good to have some magic number pre-calculated value tests as well to show that it works with only one side of the API. I.e. provide a packed bitstring with known properties and then pop the known bits out of it.

ryan-summers: We could have some weird symmetric bug (i.e. happens on both input and output) that we aren't finding because of this.

miniconf/src/packed.rs
```diff
@@ -0,0 +1,219 @@
...
+        }
+    }
+}
+
+impl Keys for Packed {
+    type Item = usize;
+
+    #[inline]
+    fn next(&mut self, len: usize) -> Option<Self::Item> {
+        self.pop_msb(Self::bits_for(len.saturating_sub(1)))
```
ryan-summers, line 189: Why this subtraction?

miniconf/src/packed.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// it is left unchanged and `None` is returned.
+    ///
+    /// # Args
+    /// * `bits`: Number of bits to pop. `bits <= Self::CAPACITY`
+    #[inline]
+    pub fn pop_msb(&mut self, bits: u32) -> Option<usize> {
+        let s = self.get();
+        if let Some(new) = Self::new(s << bits) {
+            self.0 = new.0;
+            Some((s >> (Self::CAPACITY - bits)) >> 1)
```
ryan-summers, line 156: Maybe add a note that the extra >>1 is for getting rid of the marker bit. I didn't follow this at first.

miniconf/src/postcard.rs
```diff
@@ -0,0 +1,76 @@
...
+    /// # use miniconf::{Tree, TreeKey, Postcard, Packed};
+    /// # use postcard::de_flavors::Slice;
+    /// #[derive(Tree, Default)]
+    /// struct S {
+    ///     foo: u32,
+    ///     #[tree(depth=1)]
+    ///     bar: [u16; 3],
+    /// };
+    /// let mut s = S::default();
+    /// let p = Packed::from_lsb(0b1110.try_into().unwrap());
```
ryan-summers, line 18: I'm not following this. I would expect:
1. The leading `0b1, indicates leaf tree `bar`
2. The remaining `0b110` only needs 2 bits to index all of `bar` (since there are 3 elements, 2 bits should be required)
3. `0b11` should be the index into `bar`, which would be invalid, since `S::bar[3]` is an out-of-index error.

Maybe we should explain the packed key a bit more? It's hard to wrap my head around.

Also, what's the intended use case here? I don't imagine people will actually be manually calculating these packed bitfields.

miniconf/tests/packed.rs
```diff
@@ -0,0 +1,71 @@
...
+        Settings::path(Packed::default(), &mut p, "/"),
+        Err(Error::TooShort(0))
+    );
+    p.clear();
+
+    for q in Settings::iter_paths::<String>("/") {
+        let q = q.unwrap();
+        let (a, _d) = Settings::packed(q.split("/").skip(1)).unwrap();
+        Settings::path(a, &mut p, "/").unwrap();
+        assert_eq!(p, q);
```
ryan-summers, line 26: Can you name them something more logical than `p`, `q`, `a`, and `d`? I don't know what these variables are all supposed to be.

miniconf/tests/packed.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+    println!(
+        "{:?}",
+        Settings::iter_packed()
+            .map(Result::unwrap)
+            .collect::<Vec<_>>()
+    );
+
+    assert_eq!(
+        Settings::path(Packed::new(0b01 << 29).unwrap(), &mut p, "/"),
```
ryan-summers, line 37: Not sure what this is testing? Is this checking the returned depth?

ryan-summers (commented):
I really like the update to `Keys` since it makes more sense than `Indices`, but I'm not fully understanding the use case here for `Packed`. Why are we exposing this packed bit representation to users? It seems extremely opaque to be able to figure out the bit representation without knowing the keys in advance and having code calculate it all for you.

miniconf/src/packed.rs
jordens: Maximum index is `len - 1`.

miniconf/src/postcard.rs
jordens: Docs for `Packed` are:

https://github.com/quartiq/miniconf/blob/f04a851878968107df92abe6e866538b3eb65547/miniconf/src/packed.rs#L7-L30

I'll spell out the example.

miniconf/tests/packed.rs
jordens: Yes. It's asserting that `path()` succeeds with the correct depth.

miniconf/tests/packed.rs
jordens: done

miniconf/src/packed.rs
jordens: Done in the struct doctest example.

miniconf/src/packed.rs
jordens: It's to avoid shift wrap. Comment added.

miniconf/src/postcard.rs
jordens: Done. I'll spell out a use case in the Packed struct doc.

miniconf/src/iter.rs
```diff
@@ -20,30 +19,89 @@ pub struct PathIter<'a, M: ?Sized, const Y: usize, P> {
...
+            }
+            // Non-root leaf (depth @ 1..): advance index at current depth
+            Ok(depth) => {
+                self.count = self.count.map(|c| c - 1);
+                self.state[depth - 1] += 1;
+                State::Leaf(depth)
+            }
+            Err(Error::Inner(e)) => State::Err(e),
+            // Note(unreachable):
+            // * NotFound(0): Not having consumed any name/index, the only possible case
```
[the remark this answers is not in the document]
jordens: Done

jordens:
> I really like the update to `Keys` since it makes more sense than `Indices`

I don't think there ever was `Indices`. You mean `Iterator<Item: Keys>`?

ryan-summers:
> I don't think there ever was `Indices`. You mean `Iterator<Item: Keys>`?

I was referring primarily to using `usize` slices to index into things. It makes more sense to generically iterate across the tree using `Keys`, which is then implemented directly for `usize` slices if that's how the user wants to index.

miniconf/src/packed.rs
```diff
@@ -0,0 +1,249 @@
...
+///
+/// let mut p = Packed::EMPTY;
+/// let mut p_lsb = 1; // marker
+/// for (bits, value) in [(2, 3), (1, 0), (0, 0), (3, 5)] {
+///     p.push_lsb(bits, value).unwrap();
+///     p_lsb <<= bits;
+///     p_lsb |= value;
+/// }
+/// assert_eq!(p_lsb, 0b1_11_0__101);
+/// assert_eq!(p, Packed::from_lsb(p_lsb.try_into().unwrap()));
```
ryan-summers, line 51: Why does the user have to specify the marker when providing an LSB-aligned version? It seems like this could be implicitly understood to exist above the MSB of the bit representation of the data.

I think this is what was tripping me up previously - I expected the marker bit to be handled entirely internally.

jordens:
> I was referring primarily to using usize slices to index into things. It makes more sense to generically iterate across the tree using Keys, which is then implemented directly for usize slices if that's how the user wants to index.

That was also possible before (any `Iterator<Item: Key>` would work). Just that `Keys` has a slightly different API than `Iterator` since it passes `len` as well which is needed for `Packed`.

miniconf/src/packed.rs
ryan-summers: Aaahhh yeah I forgot we could have leading zeros in an LSB-aligned implementation. That makes sense, thanks.

ryan-summers (approved):

PR #201, opened 2024-04-21
setter/getter revamp

jordens:
This introduces some changes to internal setter callbacks and identifies a nice clean use case for them. Also introduces `typ` derive macro attribute.

- add TreeKey::len()
- PhantomData: TreeKey passthrough
- back to y-1 PhantomData
- rm PhantomData
- use internal setter like get_mut()


miniconf/tests/validate.rs
```diff
@@ -36,26 +37,63 @@ impl Settings {
...
+    assert_eq!(s.set_json("/v", b"-1.0"), Err(Error::InvalidLeaf(1, "")));
     assert_eq!(s.v, 1.0); // remains unchanged
-    s.set_json("/i/a", "1.0".as_bytes()).unwrap();
+    s.set_json("/i/a", b"1.0").unwrap();
     assert_eq!(s.i.a, 1.0);
-    assert!(matches!(
-        s.set_json("/i/a", "-1.0".as_bytes()),
-        Err(Error::InvalidInternal(1, _))
-    ));
+    assert_eq!(s.set_json("/i/a", b"-1.0"), Ok(4));
```
ryan-summers, line 54: Shouldn't we add a test that it now fails after this occurs because i.a is now negative?

ryan-summers (changes requested):
One request to update a test, otherwise looks good to me. This certainly looks like a more useful case for non-leaf setters.

ryan-summers (approved):

PR #210, opened 2024-04-29
lookup

jordens:
- add accessor examples
- factor TreeLookup

### DIFF
--- a/miniconf/src/array.rs
+++ b/miniconf/src/array.rs
@@ -2,54 +2,50 @@ use core::any::Any;
 
 use crate::{
     digits, increment_result, Error, Keys, Metadata, Traversal, TreeAny, TreeDeserialize, TreeKey,
-    TreeSerialize,
+    TreeLookup, TreeSerialize,
 };
 use serde::{de::Deserialize, Deserializer, Serialize, Serializer};
 
-fn get<'a, const Y: usize, const N: usize, K, T>(
+fn get<'a, const N: usize, K, T>(
     arr: &'a [T; N],
     keys: &mut K,
+    drain: bool,
 ) -> Result<&'a T, Traversal>
 where
-    [T; N]: TreeKey<Y>,
+    [T; N]: TreeLookup,
     K: Keys,
 {
-    let index = keys.next::<Y, [T; N]>()?;
+    let index = keys.next::<[T; N]>()?;
     let item = arr.get(index).ok_or(Traversal::NotFound(1))?;
-    if Y == 1 {
-        keys.finalize::<1>()?;
+    if drain && !keys.is_empty() {
+        Err(Traversal::TooLong(1))
+    } else {
+        Ok(item)
     }
-    Ok(item)
 }
 
-fn get_mut<'a, const Y: usize, const N: usize, K, T>(
+fn get_mut<'a, const N: usize, K, T>(
     arr: &'a mut [T; N],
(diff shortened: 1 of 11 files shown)


miniconf/src/packed.rs
```diff
@@ -151,26 +162,36 @@ impl Packed {
...
-        // Note(unwrap): we ensure there is at least the marker bit set
-        NonZeroUsize::new(((self.get() >> 1) | (1 << Self::CAPACITY)) >> self.trailing_zeros())
-            .unwrap()
+    pub const fn into_lsb(self) -> NonZeroUsize {
+        match NonZeroUsize::new(
+            ((self.0.get() >> 1) | (1 << Self::CAPACITY)) >> self.0.trailing_zeros(),
+        ) {
+            Some(v) => v,
+            // We ensure there is at least the marker bit set
+            None => unreachable!(),
```
ryan-summers, line 171: Why not unwrap instead? Not sure why this is being changed

miniconf/src/key.rs
```diff
@@ -1,53 +1,81 @@
-use crate::{Traversal, TreeKey};
+use crate::Traversal;
+
+/// The capability to look up top level field names and convert to indices
+///
+/// This trait is derived together with [`crate::TreeKey`].
+pub trait TreeLookup {
```
ryan-summers, line 6: I was expecting to find this in `tree.rs` given the name of the trait

miniconf/src/array.rs
```diff
@@ -101,32 +97,34 @@ macro_rules! depth {
...
                 item.get_mut_by_key(keys).map_err(Traversal::increment)
             }
         }
     )+}
 }
 depth!(2 3 4 5 6 7 8 9 10 11 12 13 14 15 16);
 
-// Y == 1
-impl<T, const N: usize> TreeKey for [T; N] {
+impl<const N: usize, T> TreeLookup for [T; N] {
```
ryan-summers, line 116: Why is the order of params swapped from below? I thought semantically const-generics had to come last?

ryan-summers (commented):
Some minor questions/cleanup comments.

miniconf/src/packed.rs
jordens: not const

miniconf/src/key.rs
jordens: Interestingly this turned out to not be much like the other traits in `tree`. `TreeLookup` isn't depth-parametric and it's strictly only single (top) level len and name_to_index.

miniconf/src/array.rs
jordens: I don't think so. The impl order of const generics and types didn't seem to matter. And it shows up nowhere else.

miniconf/src/key.rs
ryan-summers: Yeah I noticed that as well. Just mainly threw me off because it has the `Tree` prefix on the trait name.

miniconf/src/packed.rs
ryan-summers: Ah okay, thanks!

miniconf/src/key.rs
jordens: Ack. Let's call it `KeyLookup`.

ryan-summers (approved):

PR #211, opened 2024-04-30
release v0.11.0

jordens:
- v0.11.0 release prep
- readme, `get{_mut}_by_key` -> `{ref,mut}_any_by_key`
- names, readme, add `{ref,mut}_by_key` default
- support limited depth iteration
- docs/readme
- bevy_reflect comparison
- minimq bump


miniconf/src/array.rs
```diff
@@ -93,30 +93,29 @@ macro_rules! depth {
...
     )+}
 }
 depth!(2 3 4 5 6 7 8 9 10 11 12 13 14 15 16);
 
 impl<const N: usize, T> KeyLookup for [T; N] {
-    fn len() -> usize {
-        N
-    }
+    const LEN: usize = N;
+    const NAMES: &'static [&'static str] = &[];
```
ryan-summers, line 118: Whats the point of adding these to the keylookup trait? Not seeing the value of the `NAMES` const.

jordens: It's cleaner. Otherwise in the derive macro we need to try to hide it in the inherent impl.

PR #215, opened 2024-07-01
Refactoring to use aiomqtt python client

### LINKED ISSUE #205: py: Migrate to `aiomqtt`
Looks more "idiomatic" ...
Let's explore it.

https://github.com/sbtinstruments/aiomqtt

ryan-summers:
Fixes #205 by migrating to aiomqtt


py/miniconf-mqtt/miniconf/miniconf.py
```diff
@@ -43,103 +41,99 @@ def __init__(self, client, prefix):
...
-                MiniconfException(f"Received code: {code}, Message: {response}")
-            )
-
-        del self._inflight[request_id]
-
-    async def _do(self, *args, **kwargs):
+            async with Client(self.broker, protocol=MQTTv5, logger=LOGGER) as client:
+                await client.subscribe(self.response_topic)
+                LOGGER.info("MQTT listener started")
+                async for message in client.messages:
```
ryan-summers, line 55: For some odd reason, this cannot use `self.client`. If you try, the `async for` loop immediately raises a `StopAsyncIteration` exception and the function exits. If you put it in a `while True` loop, it will never yield any messages even though it is properly subscribed.

I suspect there's something wrong about how the client is shared across the async tasks. The workaround here is to use a duplicate MQTT client that only lives in this task.

jordens: my first guess was windows but you have the selector event loop right.
so I don't know.

ryan-summers: Its a bug in aiomqtt in how they create the messages generator. If you use the `-d` flag, it causes iteration to happen in the discovery loop, but then fail later. If you don't use the `-d` flag, discovery doesn't happen (so we don't iterate messages), and iteration works later.

I think the problem is that `aiomqtt` is constructing the async iterator during client initialization as opposed to whenever `messages` is requested, so its been exhausted when we call it a second time.

ryan-summers: PR for the fix is https://github.com/sbtinstruments/aiomqtt/pull/312

jordens:
Nice. I had also started similar work. What I wanted to do while we are at it is move to pyproject.toml or whatever is en-vogue currently.

ryan-summers:
Yeah, I replaced `setup.py` with `pyproject.toml`: https://github.com/quartiq/miniconf/pull/215/files#diff-fdc4b7f0e2e856547790279b643c38f6d84f259ce01a172c2404df3245956c05

py/miniconf-mqtt/pyproject.toml
```diff
@@ -0,0 +1,20 @@
...
+build-backend = "setuptools.build_meta"
+
+[license]
+text = "MIT"
+
+[project]
+name = "miniconf-mqtt"
+# Note: keep this in sync with Cargo.toml
+version = "0.11.0"
+#url = "https://github.com/quartiq/miniconf"
```
jordens, line 12: Can't we have a url?

py/miniconf-mqtt/miniconf/__main__.py
```diff
@@ -67,36 +69,38 @@ def main():
...
+            interface = Miniconf(client, prefix)
+
+            for arg in args.paths:
+                try:
+                    path, value = arg.split("=", 1)
+                except ValueError:
+                    value = await interface.get(arg)
+                    print(f"{arg} = {value}")
+                else:
+                    print(f'Setting {arg} to {value}')
```
jordens, line 96:
```suggestion
```

Or merge with the print in L98.
Or `log::info()`

py/miniconf-mqtt/miniconf/discover.py
```diff
@@ -30,35 +33,40 @@ async def discover(
...
+    else:
+        return await do_discovery(client, prefix, rel_timeout, abs_timeout)
 
-    def on_subscribe(_client, mid, _qos, _props):
-        sub[mid].set_result(True)
 
+async def do_discovery(
+        client: Client, prefix: str, rel_timeout: float, abs_timeout: float
+    ) -> List[str]:
+    """ Do the discovery operation. Refer to `discover` doc strings for parameters. """
```
jordens, line 46:
```suggestion
    """Do the discovery operation. Refer to `discover` doc strings for parameters. """
```

py/miniconf-mqtt/miniconf/discover.py
```diff
@@ -30,35 +33,40 @@ async def discover(
...
-    t_start = asyncio.get_running_loop().time()
-    sub[client.subscribe(f"{prefix}{suffix}")] = fut
-    await fut
-    t_rtt = asyncio.get_running_loop().time() - t_start
-    await asyncio.sleep(rel_timeout * t_rtt + abs_timeout)
+    listen_task = asyncio.create_task(listen())
+    try:
+        await asyncio.wait_for(listen_task, timeout=rel_timeout * t_subscribe + abs_timeout)
+    except asyncio.TimeoutError:
+        listen_task.cancel()
```
jordens, line 68: `wait_for()` already does `cancel()`

https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for

py/miniconf-mqtt/miniconf/miniconf.py
```diff
@@ -44,41 +42,22 @@ def __init__(self, client, prefix):
...
-            return
+        self.subscribed = False
 
-        self._pending_subscriptions[mid].set_result(True)
-        del self._pending_subscriptions[mid]
 
-    def _on_message(self, _client, topic, payload, _qos, properties):
-        """Callback function for when messages are received over MQTT.
+    def process_message(self, message: Message):
+        """ Process a received MQTT message. """
```
jordens, line 50:
```suggestion
        """Process a received MQTT message."""
```

And elsewhere.

py/miniconf-mqtt/miniconf/miniconf.py
```diff
@@ -44,41 +42,22 @@ def __init__(self, client, prefix):
...
+        try:
+            properties = message.properties.json()
+        except AttributeError:
+            properties = {}
+
+        LOGGER.debug(f"Received {message.topic}: {message.payload} [{properties}]")
```
jordens, line 56: IIRC one reason behind the f-string lint for logging is that it is expensive compared to lazy formatting even if the message ends up being discarded.
I'd like to keep the high-traffic situation in mind.

py/miniconf-mqtt/miniconf/miniconf.py
```diff
@@ -125,20 +104,45 @@ def _on_message(self, _client, topic, payload, _qos, properties):
 
         del self._inflight[request_id]
 
-    async def _do(self, *args, **kwargs):
+
+    async def _process_until_complete(self, fut):
```
jordens, line 108: Does this work if there are multiple concurrent commands? Doesn't the (common?) message async iterator collide?

py/miniconf-mqtt/miniconf/__main__.py
ryan-summers: I'd rather not mix in `log` here because there's a lot of MQTT traffic that gets logged if you go down to `info`

py/miniconf-mqtt/miniconf/miniconf.py
ryan-summers: The underlying iterator yields from an `asyncio.Queue` of incoming messages. That queue would only yield a single message to a single consumer, which would then process it and finish any pending futures for other tasks.

In summary, I don't think there's any issue with multiple in-flight messages / commands here.

py/miniconf-mqtt/miniconf/discover.py
ryan-summers: Fixed in 51d73266a37342b6646bcd9aafaab9c4b9edb983

py/miniconf-mqtt/pyproject.toml
ryan-summers: I simply forgot to fix this after initially commenting it out. Added the proper URLs section for the package in 51d73266a37342b6646bcd9aafaab9c4b9edb983

py/miniconf-mqtt/miniconf/miniconf.py
ryan-summers: Fixed in f4de72322c26cebfb5b53bd81172b1a8138c0f81

py/miniconf-mqtt/miniconf/miniconf.py
jordens: AFAICT it doesn't work nicely with concurrent commands at all. The queue will dish out each received message just once to whatever task is up on its `get()`. That means with this architecture should be different. There should be just exactly one task iterating over messages and doing the dispatch into `_inflight`.
Create that task at the start. Don't wait for it.

jordens (changes requested):
https://github.com/quartiq/miniconf/pull/215#discussion_r1664356736

py/miniconf-mqtt/miniconf/miniconf.py
ryan-summers: The futures are still stored in the `_inflight` dictionary. So when _any_ task processes a received MQTT message, the UUID attached to the message indexes into the `_inflight` dictionary and then marks the future as complete. If the just-processed message did _not_ handle the current tasks's future (i.e. it processed a response for a different `get()`/`set()`), then it will `await` the next message because its future is still pending. When this happens, the previous task `get()`/`set()` will then poll its future, notice that it has been marked as complete, and will return normally.

This seems like it would all work exactly as expected from a multiple-inflight perspective.

jordens: At the very least it's needlessly creating tasks. Just one `listen()` task for the lifetime of the client is sufficient.

jordens: Creating the listen task might fit nicely with reorg the `subscribed` flag and the initial subscription (which is what causes the the messages to be sent/received). Maybe make that async subscribe/unsubscribe explicit part of the API? Or sync create the `subscribe+async iter` task in init and provide a `close()`?

jordens:
@ryan-summers I can't request your review, but let's have a look.

jordens (approved):

py/miniconf-mqtt/miniconf/discover.py
```diff
@@ -30,35 +33,30 @@ async def discover(
...
     t_start = asyncio.get_running_loop().time()
-    sub[client.subscribe(f"{prefix}{suffix}")] = fut
-    await fut
-    t_rtt = asyncio.get_running_loop().time() - t_start
-    await asyncio.sleep(rel_timeout * t_rtt + abs_timeout)
-
-    client.unsubscribe(f"{prefix}{suffix}")
-    client.on_subscribe = lambda *_a, **_k: None
-    client.on_message = lambda *_a, **_k: None
+    await client.subscribe(topic)
```
ryan-summers, line 41: We still have `client` type variables as `str | Client`, but its now only `Client`

ryan-summers: Changed to just `Client` in 095c23edab750f2e4946c71d1e8065537472ccde

py/miniconf-mqtt/miniconf/__main__.py
```diff
@@ -5,20 +5,33 @@
...
-from gmqtt import Client
+from aiomqtt import Client
+import paho.mqtt
 
 from .miniconf import Miniconf, MiniconfException
 from .discover import discover
 
+MQTTv5 = paho.mqtt.enums.MQTTProtocolVersion.MQTTv5
+
+if sys.platform.lower() == "win32" or os.name.lower() == "nt":
```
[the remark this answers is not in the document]
jordens: Choosing the event loop should definitely be done not by the library but by the downstream app. Imagine you import miniconf somewhere deep within code. That must not mess with the reactor.
As annoying as it is, this needs to be replicated in each user site (that wants to use windows).

PR #216, opened 2024-07-01
Updating CI workflow

ryan-summers:
This PR updates CI to address deprecation notices and false-positives.


jordens (commented):
all good. just need to address crosstrait nightly errors.
I think we want to move crosstrait to its own repo.

jordens:
I'm fine with force-landing this now and then #217 to move crosstrait.

PR #217, opened 2024-07-02
CLI/NodeIter/Transcode

jordens:
### Changed

* `{Path,Indices,Packed}Iter` -> `NodeIter`
* `TreeKey::iter_{paths,indices,packed}` -> `TreeKey::nodes`
* `TreeKey::{path,indices,packed,json_path}` -> `TreeKey::transcode`/`Transcode::transcode`
* `crosstrait` now has its [own repository](https://github.com/quartiq/crosstrait)
* `Keys::is_empty()` -> `Keys::finalize()`
* `traverse_by_key` ensures `Keys::finalize()`
* `NodeIter::count()` -> `NodeIter::exact_size()` to disambiguate from `Iterator::count()`

### Added

* `Transcode` trait for Keys transcoding and node lookup.
* `Transcode` and `NodeIter` now return `Node` with `NodeType` information (leaf or internal).
* `Keys::chain` and `Chain` to concatenate two `Keys` of different type.
* `miniconf_cli`: a menu/command line interface
* `Path`, `JsonPath`/`JsonPathIter`, `Indices`, `KeysIter` wrapper types for more ergonomic/succinct
  `Transcode`/`IntoKeys`/`Keys` handling

### Removed

* `digits()` gone in favor of using `usize::checked_ilog10()`

For whatever reason this saves about 5 kB on stabilizer/dual-iir and fls


miniconf/src/key.rs
```diff
@@ -54,33 +54,57 @@ impl Key for &str {
...
+        T::next::<M>(self)
+    }
+
+    fn finalize(&mut self) -> bool {
+        T::finalize(self)
     }
 }
 
-/// Capability to be converted into a [`Keys`]
+/// Be converted into a `Keys`
```
ryan-summers, line 107: Why are we using this custom `IntoKeys` trait instead of the bound `core::convert::Into<Keys>`? Is it because `Keys` doesn't have a known type?

miniconf/src/node.rs
```diff
@@ -0,0 +1,263 @@
...
+            Err(Error::Inner(depth, ())) => Err(Traversal::TooShort(depth)),
+            Err(Error::Traversal(err)) => Err(err),
+            Err(Error::Finalization(())) => unreachable!(),
+        }
+    }
+}
+
+/// Look up an `IntoKeys` in a `TreeKey` and transcode into `Self`.
+pub trait Transcode {
+    /// Perform a lookup and transcode the keys into self
```
ryan-summers, line 106:
```suggestion
/// Look up an `IntoKeys` in a `TreeKey` and transcode into a `Node`.
pub trait Transcode {
    /// Perform a lookup and transcode the keys into a node
```

miniconf/src/node.rs
```diff
@@ -0,0 +1,263 @@
...
+    ///
+    /// Returns node type and depth info.
+    ///
+    /// `Err(Traversal::TooShort(depth))` indicates that `Self` does not have sufficient
+    /// capacity and failed to encode the key at the given depth.
+    fn transcode<M, const Y: usize, K>(&mut self, keys: K) -> Result<Node, Traversal>
```
ryan-summers, line 112: I'm personally not a huge fan of this trait and function name. It took me a long time to figure out what they were doing before I got to the definitions here.

Maybe something like `Lookup` instead? Not sure. Not a deal breaker, just a comment.

miniconf/src/node.rs
```diff
@@ -0,0 +1,263 @@
...
+impl<T, const S: char> From<T> for Path<T, S> {
+    fn from(value: T) -> Self {
+        Path(value)
+    }
+}
+
+impl<'a, T: AsRef<str> + ?Sized, const S: char> IntoKeys for &'a Path<T, S> {
+    type IntoKeys = KeysIter<Skip<Split<'a, char>>>;
+    fn into_keys(self) -> Self::IntoKeys {
+        self.0.as_ref().split(self.separator()).skip(1).into_keys()
```
ryan-summers, line 179: Doesn't this assume that the path begins with the separator? If the path didn't, wouldn't we skip the root element?

miniconf_cli/src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+
+    #[tokio::test]
+    async fn all() {
+        let mut buf = [0; 1024];
+        let mut s = Set::default();
+        let mut stdout = embedded_io_adapters::tokio_1::FromTokio::new(tokio::io::stdout());
+        let mut menu = Menu::<Set, Y>::default();
+        s.c = Some(8);
+        menu.enter("b/0").unwrap();
+        menu.set(&mut s, b"1234").unwrap();
```
ryan-summers, line 293: Shouldn't we assert that `s.b[0] == 1234`?

miniconf_cli/src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+        menu.exit(2).unwrap();
+        menu.push("f/1/e").unwrap().0.set(&mut s, b"9").unwrap();
```
ryan-summers, line 295: Same, shouldn't we check the original struct and verify it was set properly?

ryan-summers (approved):

miniconf/src/node.rs
ryan-summers: Nevermind, transcode makes sense for what this is doing (converting between key/path types and formats)

miniconf/src/node.rs
ryan-summers: I was confused reading these docs about what is being transcoded. It makes sense after the review that we're transcoding into other key types - we should specifically mention that instead of an ambiguous `Self` type of the trait implementer

miniconf/src/key.rs
jordens: Rationale is the same as for `IntoIterator`. One reason is indeed to have `Keys` as an associated type and not as an generic type.

miniconf/src/node.rs
jordens: The rationale is as before. `Path`s start with the separator (as documented). The issue is `"".split('/') -> [""]`: disambiguate between e.g. `Option<T>: TreeKey<1>` which does not consume a key and the empty name, which consumes a key.

miniconf_cli/src/lib.rs
jordens: Yeah. I presume `miniconf_cli` will change while we try to use it as the base for `serial_settings` and these tests will get some more refactoring iterations.

miniconf_cli/src/lib.rs
jordens: These were more playground to see how it'll look that actual unittests. Hopefully they'll improve .. ;)

miniconf/src/node.rs
jordens: #218

PR #218, opened 2024-07-03
docs

jordens:
- **introduce custom .split().skip(1) impl**
- **refine Transcode docs**

### DIFF
--- a/miniconf/src/node.rs
+++ b/miniconf/src/node.rs
@@ -1,9 +1,8 @@
 use core::{
     fmt::Write,
-    iter::{Copied, Skip},
+    iter::Copied,
     ops::{Deref, DerefMut},
     slice::Iter,
-    str::Split,
 };
 
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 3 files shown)


miniconf/src/node.rs
```diff
@@ -173,10 +173,45 @@ impl<T, const S: char> From<T> for Path<T, S> {
...
+
+impl<'a, const S: char> Iterator for PathIter<'a, S> {
+    type Item = &'a str;
+
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        self.0.map(|s| {
+            let pos = s
+                .chars()
+                .map_while(|c| (c != S).then_some(c.len_utf8()))
```
ryan-summers, line 204: Would https://doc.rust-lang.org/core/iter/trait.Iterator.html#method.skip_while work for us? I.e. just skip the empty string?

ryan-summers: Ah I think the docs about about empty keys vs empty path for options above explains it.

jordens: I think skip while empty would work but it's less strict than the  implicit definition above (a path element is always prefixed by a separator). And we'd have to be much kore careful w.r.t. composability in e.g. CLI, Mqtt, and serial settings where then we'd have to conditionally insert the separator if we allow path to start without.

ryan-summers: Yeah being strict here feels more right for sure. Thanks for humoring me

miniconf/src/node.rs
```diff
@@ -173,10 +173,47 @@ impl<T, const S: char> From<T> for Path<T, S> {
...
+                .sum();
+            let (left, right) = s.split_at(pos);
+            self.0 = right.get(S.len_utf8()..);
```
ryan-summers, line 207: How is this semantically different than `split(S).skip(1)`?

jordens: `PathIter` has exactly the same Iterator behavior as `Skip<Split<_>>`. It's just 2 kB less code on stabilizer.
https://github.com/quartiq/miniconf/blob/0cd1773c192d14ad5423595d7817fda9df161785/miniconf/src/node.rs#L306-L314

ryan-summers: Huh, interesting. Alright, seems fine to me.

ryan-summers (approved):

PR #221, opened 2024-07-09
follow up changes to mqtt rework

jordens:


py/test.sh
```diff
@@ -11,12 +12,17 @@ cargo build -p miniconf_mqtt --example mqtt
...
-python -m miniconf -b localhost -d dt/sinara/dual-iir/+ '/afe/0="G1"'  # SET
-python -m miniconf -b localhost -d dt/sinara/dual-iir/+ '/afe/0='      # CLEAR
-python -m miniconf -b localhost -d dt/sinara/dual-iir/+ '/afe?' '?'    # LIST-GET
-python -m miniconf -b localhost -d dt/sinara/dual-iir/+ '/afe!'        # DUMP
-sleep 1  # dump is asynchronous
+# GET SET CLEAR LIST DUMP
+$MC '/afe/0' '/afe/0="G10"' '/afe/0=' '/afe?' '?' '/afe!'
+sleep 1  # DUMP is asynchronous
+
+$MC '/four=5'
```
ryan-summers, line 22: Shouldn´t we test that this failed? I.e. use the same `test $? -ne 0 && exit 1`?

jordens: The `set -e` does that for everything.

ryan-summers: Ah sorry, I got confused and thought the test was for <4, not >4

ryan-summers (approved):

PR #225, opened 2024-07-10
alive payload

jordens:
- **TreeKey: fix docs on traverse/node, doctest updates**
- **mqtt: support alive info**
- **make alive configurable late**

### DIFF
--- a/miniconf_mqtt/src/lib.rs
+++ b/miniconf_mqtt/src/lib.rs
@@ -213,7 +213,6 @@ impl From<ResponseCode> for minimq::Property<'static> {
 /// let mut client = miniconf_mqtt::MqttClient::new(
 ///     std_embedded_nal::Stack::default(),
 ///     "quartiq/application/12345", // prefix
-///     "1", // alive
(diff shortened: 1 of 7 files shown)


miniconf_mqtt/src/lib.rs
```diff
@@ -275,10 +274,20 @@ where
...
         })
     }
 
+    /// Set the payload published on the `/alive` topic when connected to the broker.
+    ///
+    /// The default is to publish `1`.
+    /// The message is retained by the broker.
+    /// On disconnect the message is cleared retained through an MQTT will.
+    pub fn set_alive(&mut self, alive: &'buf str) {
+        self.alive = alive;
```
ryan-summers, line 288: Don't we want to put the state machine into a state where the alive message gets republished (in case a user calls this after the alive publication was already sent)

jordens: I thought about it. But I'd rather not force it.
Let's do a separate fn for reset.

ryan-summers (approved):

PR #232, opened 2024-08-28
examples

jordens:
- **write validator without closure, add example**
- **macro: avoid multiple parsing for `derive(Tree)`**
- **macro: consume mutable tree**
- **miniconf: add cli example**
- **miniconf_cli -> miniconf_menu**
- **factor examples common**
- **remove miniconf_menu, add example**
- **examples/menu: make test**


jordens:
@ryan-summers ping

ryan-summers (approved):

Issue #166, opened 2023-07-31
derive macro for complex enums


jordens:
would only change their variant at runtime outside `Miniconf`, i.e. be like option, untagged enum
They would implement exactly one depth: the max of the inner types.

MathiasKoch:
﻿I started an implementation of general enum support, but I think we might need to discuss slightly how it should work :p

### Basic enum keys

For instance, a simple unit test of the node iteration:

```rust

#[derive(PartialEq, Debug, Clone, Default, Tree)]
struct Inner {
    data: u32,
}

#[derive(Debug, Clone, Tree)]
enum Either {
    A
    B(Inner)
}

#[derive(Debug, Clone, Tree)]
enum EitherDeep {
    A
    #[tree(depth = 1)
    B(Inner)
}

#[test]
fn iterate_keys() {
    let mut expected_keys = HashSet::from(["/A", "/B"]);

    for node in Either::nodes::<Path<String, '/'>>() {
        let (path, node) = node.unwrap();
        assert!(node.is_leaf());

        assert!(expected_keys.remove(path.as_str()));
    }
    assert!(expected_keys.is_empty());
}


#[test]
fn iterate_keys_deep() {
    let mut expected_keys = HashSet::from(["/A", "/B/data"]);

    for node in EitherDeep::nodes::<Path<String, '/'>>() {
        let (path, node) = node.unwrap();
        assert!(node.is_leaf());

        assert!(expected_keys.remove(path.as_str()));
    }
    assert!(expected_keys.is_empty());
}
```

Intuitively, I would expect both of these tests to pass. 

But what happens when we want to support all forms of enums (tree attr removed for simplicity):

```rust
#[derive(Debug, Clone, Tree)]
enum EitherDeep {
    A,
    B(Inner),
    C(Inner, Inner),
    D { value: usize }
}
```

Maybe the more correct way would be to have this result in keys: `["/A", "/B/0/data", "/C/0/data", "/C/1/data", "/D/value"]`, assuming that would be possible with the current traits?

That way it should also be super simple to reuse the same code for supporting tuples, in the same manner as arrays?

The issue is, that now enums of the format `B(Inner)` has a different key scheme than a newtype struct, and Option would have, as I would expect both of these to be "transparent" (as is already the case with Option)

Obviously, another option is to only allow a single unnamed field in the `B` cases, and if users need more fields, they would need to wrap it in a struct or tuple? Though I would personally think it would be nicer to design for the full syntax support eventually?  

[...]

PR #233, opened 2024-09-23
derive support for enums with newtype/unit variants

jordens:
* Only newtype and unit variants supported currently. Non-newtype tuple structs can be used with `skip` fields. Both record structs and non-newtype tuple structs would require creating proxy structs on the fly to implement `KeyLookup` and `TreeKey` on them.
* The `.0` newtype hierarchy level is flattened intrinsically.
* Unit variants are supported by being equivalent to `skip`. They would only become relevant once there is some instrinsic support for switching variants.
* No intrinsic support for variant read. An external getter can be used.
* No intrinsic support for variant write. An external setter can be used. This needs some user code in any case since there is no `Default` per variant.
* The implementation differs from the one proposed in #166 in that even for `depth = 1`, there is no intrinsic support for variant switching.
* No support for variant `flatten`ing. This is the only thing that would be missing to support `Option`-style behavior through the derive macros.


jordens:
@MathiasKoch does this help you? There are some examples in `tests/enum.rs`.

MathiasKoch:
This looks amazing! 
I never got around to implementing it myself yet, but this will absolutely help!

